PR #9, opened 2024-06-04
Extend resources and verbs for service account `mintmaker-controller-manager`

qixiang:
There are more resources and verbs are required.


FernandesMF (approved):
Looks good to me!

PR #10, opened 2024-06-06
Run controller pod with readonly root filesystem

qixiang:
The controller pod has two containers:
* kube-rbac-proxy
* manager

These containers should run with readonly root filesystem.

This fixes the kube-linter reported errors in https://github.com/redhat-appstudio/infra-deployments/actions/runs/9369177927/job/25842546983?pr=3798
```
...
"text": "container \"manager\" does not have a read-only root file system\nobject: mintmaker/mintmaker-controller-manager apps/v1, Kind=Deployment"
...
"text": "container \"kube-rbac-proxy\" does not have a read-only root file system\nobject: mintmaker/mintmaker-controller-manager apps/v1, Kind=Deployment"
...
```

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -15,6 +15,7 @@ spec:
           capabilities:
             drop:
             - "ALL"
+          readOnlyRootFilesystem: true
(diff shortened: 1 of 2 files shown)


staticf0x (approved):
LGTM

PR #12, opened 2024-06-11
Only handle new DependencyUpdateCheck CRs in mintmaker namespace

qixiang:
* The controller only watches for CRs created in mintmaker namespace
* The controller doesn't process CRs which have been processed before, when a CR is processed, an annotation will be added to the CR.

CWFHEALTH-3083

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -27,10 +28,12 @@ import (
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 
 	mmv1alpha1 "github.com/konflux-ci/mintmaker/api/v1alpha1"
+	. "github.com/konflux-ci/mintmaker/pkg/common"
 	"github.com/konflux-ci/mintmaker/pkg/git"
 	"github.com/konflux-ci/mintmaker/pkg/k8s"
 	"github.com/konflux-ci/mintmaker/pkg/renovate"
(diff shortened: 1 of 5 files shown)


internal/controller/dependencyupdatecheck_controller_test.go
```diff
@@ -126,12 +126,36 @@ var _ = Describe("DependencyUpdateCheck Controller", func() {
...
-			createDependencyUpdateCheck(dependencyUpdateCheckKey)
+			createDependencyUpdateCheck(dependencyUpdateCheckKey, false)
 			Eventually(listJobs).WithArguments(MintMakerNamespaceName).WithTimeout(timeout).Should(HaveLen(1))
 
 			deleteComponent(componentKey)
 			deleteDependencyUpdateCheck(dependencyUpdateCheckKey)
 			deleteSecret(renovateSecretKey)
 		})
+
+		It("It should not trigger job for DependencyUpdateCheck CR which has been processed before", func() {
```
gnaponie, line 137: What does "It" stands for? I'm assuming is related to testing and to introduce a new test. It's very hard to google it, since it's a too common word :)

gnaponie:
Based on my limited experience with Go and controllers it looks good to me :)

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -67,6 +70,33 @@ func (r *DependencyUpdateCheckReconciler) Reconcile(ctx context.Context, req ctr
...
+	dependencyupdatecheck := &mmv1alpha1.DependencyUpdateCheck{}
+	err := r.client.Get(ctx, req.NamespacedName, dependencyupdatecheck)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+	}
+
+	// If the DependencyUpdateCheck has been handled before, skip it
+	if _, processed := dependencyupdatecheck.Annotations[MintMakerProcessedAnnotationName]; processed {
```
FernandesMF, line 82: I think this would read more easily if the function call and the if-clause were separated in two lines:
``` go
_, processed := dependencyupdatecheck.Annotations[MintMakerProcessedAnnotationName]
if  processed {
    ...
}
```

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -67,6 +70,33 @@ func (r *DependencyUpdateCheckReconciler) Reconcile(ctx context.Context, req ctr
...
+		log.Info(fmt.Sprintf("DependencyUpdateCheck has been processed: %v", req.NamespacedName))
```
FernandesMF, line 83: I think it would be nice to mention that the DependencyUpdateCheck will be skipped in this case. Perhaps a log message as:
```go
"DependencyUpdateCheck has already been processed: %v; skipping."
```

internal/controller/dependencyupdatecheck_controller_test.go
FernandesMF: From what I saw in the ginkgo docs, the examples they provide do not repeat "it" at the beginning of the strings:
```go
It("extracts the last name", func() { ... }
It("extracts the first name", func() { ... }
```
Perhaps leaving the "it" at the beginning of the string would make for less-clear messages in the tests? What do you think?

FernandesMF (approved):
Looks good to me! I just had a few minor suggestions.

internal/controller/dependencyupdatecheck_controller.go
qixiang: This is a common pattern in go, beside of other reasons, one important reason is the variables declared in the short statement are limited to the scope of the if block, they are not accessible outside of the if block, it reduces the chances of accidentally using these variables later in the code.
```
func main() {
	if v := true; v {
		fmt.Println("value: ", v)
	}
	fmt.Println("value: ", v) // this will raise an "undefined: v" error because v is only in scope within the if block
}
```

internal/controller/dependencyupdatecheck_controller_test.go
qixiang: Done.

internal/controller/dependencyupdatecheck_controller.go
qixiang: I didn't change this line because there are virtually no scenarios where we need to check these logs. This log entry is only relevant when the controller restarts and some CRs have already been processed but still remain in the cache.

qixiang:
Comments are addressed, so I'm going to merge this.

PR #14, opened 2024-06-19
Increase memory allocation for manager to prevent OOMKilled issue

qixiang:
The controller manager pod was consistently crashing due to being OOMKilled. Increased the memory allocation to ensure the pod has sufficient resources to operate without running out of memory.


staticf0x (approved):
LGTM

gnaponie (approved):

PR #17, opened 2024-06-21
Update client to cluster-scoped

qixiang:
The controller needs to list components in cluster, so the client can't be namespace scoped.


gnaponie (approved):

gnaponie:
looks good

PR #20, opened 2024-06-26
Allow component to have the capability to disable Mintmaker

qixiang:
This update allows component owners to disable Mintmaker for their components. To do this, add the following annotation to the component:

```
mintmaker.appstudio.redhat.com/disabled: true
```

CWFHEALTH-3162


FernandesMF (approved):
Looks good to me, nice job! 😁

PR #23, opened 2024-07-08
Fix incorrect mintmaker processed annotation check

qixiang:
Previously, the code only checked for the existence of the annotation, but did not verify its value.


gnaponie (approved):

gnaponie:
+1

PR #25, opened 2024-07-10
Use main branch when branch is missing for gitlab components

qixiang:
If revision is not explicitly specified for a GitLab component, the 'main' branch will be used as a short-term solution. In the future, we should check GitLab to determine the default branch.

### DIFF
--- a/pkg/renovate/basicauth.go
+++ b/pkg/renovate/basicauth.go
@@ -65,11 +65,17 @@ func (g BasicAuthTaskProvider) GetNewTasks(ctx context.Context, components []*gi
 						}
 						// Step 6
 						if !AddNewRepoToTasksOnTheSameHostsWithSameCredentials(tasksOnHost, component, creds) {
+							branch := component.Branch()
+							// When platform is gitlab and branch is missing, use "main" branch
+							// TODO: check Gitlab to determine the default branch
+							if component.Platform() == "gitlab" && branch == git.InternalDefaultBranch {
+								branch = "main"
+							}
 							// Step 7
 							tasksOnHost = append(tasksOnHost, NewBasicAuthTask(platform, host, endpoint, creds, []*Repository{
 								{
 									Repository:   component.Repository(),
-									BaseBranches: []string{component.Branch()},
+									BaseBranches: []string{branch},
 								},
 							}))
 						}
--- a/pkg/renovate/basicauth_test.go
+++ b/pkg/renovate/basicauth_test.go
@@ -152,6 +152,27 @@ func TestNewTasks(t *testing.T) {
 				}),
 			},
 		},
+		{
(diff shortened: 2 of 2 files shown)


pkg/renovate/basicauth.go
```diff
@@ -65,11 +65,17 @@ func (g BasicAuthTaskProvider) GetNewTasks(ctx context.Context, components []*gi
 						}
 						// Step 6
 						if !AddNewRepoToTasksOnTheSameHostsWithSameCredentials(tasksOnHost, component, creds) {
+							branch := component.Branch()
+							// When platform is gitlab and branch is missing, use "main" branch
+							// TODO: check Gitlab to determine the default branch
+							if component.Platform() == "gitlab" && branch == git.InternalDefaultBranch {
```
staticf0x, line 71: I don't really understand this. What is `git.InternalDefaultBranch`? Shouldn't it contain `main` already, at which point overwriting `branch` wouldn't be necessary? What am I missing?

qixiang: `git.InternalDefaultBranch`'s value is "$DEFAULTBRANCH": https://github.com/konflux-ci/mintmaker/blob/2b5c84ed19c684eb7dc246545b5108b02d9acea6/pkg/git/scmcomponent.go#L8

While creating a component, the revision (which is actually the branch) is optional. When MintMaker processing such component, a `ScmComponent` is created with this default value: https://github.com/konflux-ci/mintmaker/blob/2b5c84ed19c684eb7dc246545b5108b02d9acea6/pkg/git/scmcomponent.go#L18-L29

Later, Github API is used to determine the default branch for Github repositories: https://github.com/konflux-ci/mintmaker/blob/2b5c84ed19c684eb7dc246545b5108b02d9acea6/pkg/renovate/githubapp.go#L50-L52
However for Gitlab repositories, we just use `main` as default, we should also use GitLab APIs to determine the default branch later.

A lot of the controller code was copied from the build-service, and there is still a lot of redundant code that has not yet been cleaned up, so some of the code looks quite messy.

staticf0x: Thank you for the explanation. Would this work if the default branch in a repo was `master`? Could that be a case for any repo in Konflux?

qixiang: > Would this work if the default branch in a repo was `master`?

If revision/branch is not specified explicitly and the default branch is not `main`, then this won't work, Renovate will run against the `main` branch, so we need to call Gitlab API to determine the default branch later, this is currently a short-term solution.

staticf0x: Okay thank you, got it :)

staticf0x (approved):

gnaponie (approved):

gnaponie:
@qixiang do we have a card to clean up the old code from the build service?

PR #30, opened 2024-07-15
Set email ID to the currently used value to avoid Renovate not recognizing previous commits

staticf0x:
@qixiang does this make sense? Is this the correct ID?

### DIFF
--- a/pkg/renovate/githubapp.go
+++ b/pkg/renovate/githubapp.go
@@ -71,7 +71,7 @@ func newGithubTask(slug string, token string, repositories []*Repository) *Task 
 		Platform:     "github",
 		Endpoint:     git.BuildAPIEndpoint("github").APIEndpoint("github.com"),
 		Username:     fmt.Sprintf("%s[bot]", slug),
-		GitAuthor:    fmt.Sprintf("%s <123456+%s[bot]@users.noreply.github.com>", slug, slug),
+		GitAuthor:    fmt.Sprintf("%s <126015336+%s[bot]@users.noreply.github.com>", slug, slug),
 		Token:        token,
 		Repositories: repositories,
 	}


qixiang:
> @qixiang does this make sense? Is this the correct ID?

I believe it's the Github App installation ID, but not the Github APP id, so it's different for each installation, for example, you installed a Github App into your organization, and I installed the same App to my organization, they're one Github App, but two different installations.

I'm not sure what's the purpose of changing it from "123456+<github-app-slug>", could you elaborate more on this? And I'm worrying about a rare case: if the Github App is uninstalled then installed again, the id may changes.

staticf0x:
The purpose is that renovate commits under the email from the GitHub app, which is `126015336+red-hat-konflux[bot]@users.noreply.github.com` for prod, but when we create a task in MintMaker, the configured email is set to `123456+red-hat-konflux[bot]@users.noreply.github.com` which causes renovate to add the comments that it doesn't recognize the original author. But since we're using `platformCommit=true`, we somehow need to change 123456 to the correct ID on each deployment.

qixiang:
> the GitHub app, which is 126015336+red-hat-konflux[bot]@users.noreply.github.com for prod

@staticf0x where do you see the number of  `126015336`? Is it the Github App ID or an installation ID? I don't think it's the Github App id which is same across all production clusters. If this is a Github App installation ID, then it will be different for each installation.

The changed email doesn't match with what you're expecting:
```
fmt.Sprintf("%s <%s+%d[bot]@users.noreply.github.com>", slug, slug, id),
```
which will be `red-hat-konflux+<number>[bot]@users.noreply.github.com`.

I think we need to change this carefully, before the current opening PRs are having different email in their commits, if we're going to change this, could it cause Renovate can't recognize all the opening PRs?

staticf0x:
Oh you're right, I misplaced the formatting. Anyway the `126015336` comes from the commit email used by Renovate. Since we're using `platformCommit=true`, it means that Renovate is using GitHub's API to commit code into a repository. This API uses the account's email address, the one configured in GitHub, which is `126015336+red-hat-konflux[bot]@users.noreply.github.com` at least for the production. 

If you look into any Konflux repo and run `git log -1 --pretty=format:"%ae" "origin/konflux/references/main" --` (or whichever branch Renovate uses) it will output the email address used for committing. That's how I found that number. But if you look into Renovate's logs, you can see the configuration for `gitAuthor` contains the number 123456, which I think comes from MintMaker when creating the tasks. And so since `123456 != 126015336`, Renovate thinks someone else modified the PR, which leads to the notification:

> Renovate will not automatically rebase this PR, because it does not recognize the last commit author and assumes somebody else may have edited the PR.
> You can manually request rebase by checking the rebase/retry box above.

However this is directly caused by Renovate trying the rebase/retry option, so if you check that checkbox, it's gonna happen again.

So what I'm trying to achieve is to set the `gitAuthor` also to `126015336` (+ the email part of course), so that it's the same as the account's email address. But since the two email addresses come from two different places, it might be difficult to do so. My unfortunate attempt here is that I thought the app installation ID is the one we need, but now I'm not sure if it's the correct one, since there must be a difference between staging and prod at least.

This is for CWFHEALTH-3207.

staticf0x:
Maybe we'd need to query the GitHub API to get the email for the current app

qixiang:
> Maybe we'd need to query the GitHub API to get the email for the current app

I noticed in several repos (belong to different organizations), this email has been changed from:
```
123456+red-hat-konflux[bot]@users.noreply.github.com
```
to
```
126015336+red-hat-konflux[bot]@users.noreply.github.com
```
around 2024-06-25, for example in [release-service repo](https://github.com/konflux-ci/release-service.git):
```
$ git log --author konflux  --pretty=format:"%ae %aI"
...
126015336+red-hat-konflux[bot]@users.noreply.github.com 2024-06-25T16:30:52Z
123456+red-hat-konflux[bot]@users.noreply.github.com 2024-06-21T16:37:45Z
```
so this ID is same across different installations, then it shouldn't be the installation ID, it's also not the Github App id (you can get the app id in production mintmaker space, secret `pipelines-as-code-secret`), I'll check where this number come from.

staticf0x:
If this number was constant everywhere then it would be great, we could just replace it and it should work.

qixiang:
> If this number was constant everywhere then it would be great, we could just replace it and it should work.

Hi @staticf0x I haven't been able to figure out where the number in email address come from, nor how it has been changed. Could you please update this PR to just replace it with the new value `126015336`?

staticf0x:
@qixiang done, hope this works and the number doesn't change in the future

qixiang (approved):
LGTM

PR #32, opened 2024-07-18
Add renovate config and fix errors

querti:
The new directory contains the renovate config and a kustomization file to create a configmap out of it. The git path to these resources is expected to be referenced in mintmaker infra-deployments definitions. Introducing renovate config changes can be done by changing the commit SHA in mintmaker infra-deployments.

I haven't split the config into presets. The main reason is that it uses another version control system, which means that we would need to maintain versioning for the global config as well as for all of its presets. This makes things needlesly complicated and doesn't provide any benefit. Secondly, the main reason for using presets is reusing the configuration by other users. I suggest to refrain from arbitrarily splitting the config file until we know which parts other people want to reuse.

The config itself was fixed to not contain any errors and warnings. List of changes:
- Remove outer quotes from prBodyDefinitions, as this field expects a JSON object. This is the only error in the config
- Change the branchName field to branchPrefix and branchTopic. The branchName field is deprecated.
- Change '"rebaseStalePrs": true' to '"rebaseWhen": "behind-base-branch"'. This change was suggested as a renovate migration.

Refers to CWFHEALTH-3210, CWFHEALTH-3206

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,4 +4,4 @@
 # the repo. Unless a later match takes precedence, these
 # owners will be requested for review when someone opens a
 # pull request.
-*  @gnaponie @qixiang @staticf0x @FernandesMF @nvtnlucie
+*  @gnaponie @qixiang @staticf0x @FernandesMF @nvtnlucie @querti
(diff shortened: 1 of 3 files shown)


qixiang (approved):
LGTM, nice job!

PR #34, opened 2024-07-19
fix: avoid mounting non-existent ConfigMap

qixiang:
When the desired trusted CA bundle ConfigMap does not exist, MintMaker should not attempt to mount the non-existent ConfigMap.


querti (approved):
Looks good, as far as I can tell.

PR #40, opened 2024-07-25
Enable Kubernetes managers

querti:
Enable managers as grouped on this page[1]. The mintmaker image should have all the required dependencies to run these managers. Also, since they only analyze yaml/json, there shouldn't be any "hidden" dependencies either.

The new managers will use a new branch format, from renovate/\<dep name\> to konflux/mintmaker/\<branch\>/\<dep name\>. Branch name is included to resolve branch conflicts if multiple branches are onboarded onto Konflux. Mintmaker name is added so that end users can more easily understand where the PR/MR came from.

Refers to CWFHEALTH-3178

[1] https://docs.renovatebot.com/modules/manager/

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -10,7 +10,9 @@
   "requireConfig": "optional",
   "platformCommit": true,
   "autodiscover": false,
-  "enabledManagers": ["tekton", "dockerfile", "rpm", "git-submodules", "regex"],
+  "enabledManagers": ["tekton", "dockerfile", "rpm", "git-submodules", "regex", "argocd", "crossplane",
+                      "fleet", "flux", "helm-requirements", "helm-values", "helmfile", "helmsman", "helmv3",
+                      "jsonnet-bundler", "kubernetes", "kustomize"],
   "tekton": {
     "fileMatch": ["\\.yaml$","\\.yml$"],
     "includePaths": [".tekton/**"],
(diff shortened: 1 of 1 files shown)


staticf0x (approved):
Thanks for the detailed explanation, LGTM

gnaponie:
This looks nice! But you need to add `additionalBranchPrefix` and `branchPrefix` to the already enabled managers too.

qixiang:
@staticf0x @gnaponie I'd like to get more comments on using 'konflux/mintmaker' as a branch prefix. Do you think 'konflux' alone is sufficient, or should we include 'mintmaker'?

@querti do you have an example repo can be used for testing one or more of these managers to verify the branch name can take affect as expected?

querti:
@qixiang the branch name can be seen in this PR: https://github.com/querti/renovate-tests/pull/9

gnaponie:
> @staticf0x @gnaponie I'd like to get more comments on using 'konflux/mintmaker' as a branch prefix. Do you think 'konflux' alone is sufficient, or should we include 'mintmaker'?
> 
> @querti do you have an example repo can be used for testing one or more of these managers to verify the branch name can take affect as expected?

maybe should we use `konflux-mintmaker` instead of `konflux/mintmaker`?

qixiang:
> This looks nice! But you need to add `additionalBranchPrefix` and `branchPrefix` to the already enabled managers too.

When changing branch names, we need to be cautious. Existing open PRs with the old branch names can become stale and cause confusions if we use new branch names. If we're going to change branch name for existing managers we probably need to communicate this with Konflux team or users carefully.

qixiang:
> maybe should we use `konflux-mintmaker` instead of `konflux/mintmaker`?

If we decide to include mintmaker in the prefix, I'd prefer konflux/mintmake to align with the style of existing Konflux PRs. Changing the branch name is a significant change in Renovate, and we need to consider it carefully.

gnaponie:
> > This looks nice! But you need to add `additionalBranchPrefix` and `branchPrefix` to the already enabled managers too.
> 
> When changing branch names, we need to be cautious. Existing open PRs with the old branch names can become stale and cause confusions if we use new branch names. If we're going to change branch name for existing managers we probably need to communicate this with Konflux team or users carefully.

That's a fair point. But if we won't do that we won't resolve with this commit the conflict with multiple branches. We'd still need to find another solution for the already enabled managers.

qixiang:
> That's a fair point. But if we won't do that we won't resolve with this commit the conflict with multiple branches. We'd still need to find another solution for the already enabled managers.

I understand your point. This PR is not intended to resolve the issue with enabled managers. Its purpose is to add new Kubernetes managers and prevent similar branch issues with these new managers, which would be difficult to change later. For enabled managers, they will be fixed in other PRs :)

gnaponie:
> I understand your point. This PR is not intended to resolve the issue with enabled managers. Its purpose is to add new Kubernetes managers and prevent similar branch issues with these new managers, which would be difficult to change later. For enabled managers, they will be fixed in other PRs :)

:+1:  let's do it then. I don't like that much that we have inconsistent behaviors across different managers. But perhaps that's "the lesser evil" in this situation :)

qixiang:
@gnaponie However, this reminds me that we have a new manager `git-submodules`, which hasn't been deployed to staging or production yet. We should change the branch now rather than waiting until after deployment.

gnaponie:
Yes, @querti could you add this options for `git-submodules` too? Then we can merge this PR. Thank you.

querti:
> Yes, @querti could you add this options for `git-submodules` too? Then we can merge this PR. Thank you.

Done.

qixiang (approved):
Seems there is no objections to branch prefix of "konflux/mintmaker/", LGTM.

PR #58, opened 2024-08-27
Merge registry pull secrets and mount them to the job pod

staticf0x:
One caveat is that the newly created secret cannot be created as base64 encoded string and I don't know why. It works with plain text. It's this line:
```go
encodedMergedConfig := base64.StdEncoding.EncodeToString(mergedConfigJson)
```

### DIFF
--- a/pkg/renovate/job.go
+++ b/pkg/renovate/job.go
@@ -2,6 +2,8 @@ package renovate
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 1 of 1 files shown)


pkg/renovate/job.go
```diff
@@ -112,6 +201,8 @@ func (j *JobCoordinator) Execute(ctx context.Context, tasks []*Task) error {
 		return nil
 	}
 
+	registry_secret, err := j.createMergedPullSecret(ctx)
```
qixiang, line 204: This secret will live there forever, we need to set this secret to be owned by the job, then when the job is deleted automatically, this secret will be deleted as well.

For example: https://github.com/konflux-ci/mintmaker/blob/8dde1c20a51bca9b74dabb42d7defc703b4432bd/pkg/renovate/job.go#L232-L237

pkg/renovate/job.go
```diff
@@ -146,6 +237,20 @@ func (j *JobCoordinator) Execute(ctx context.Context, tasks []*Task) error {
 								},
 							},
 						},
+						{
+							Name: "registry-secrets",
```
qixiang, line 241: When there is no such secret created (e.g. we don't have any pull secret prepared in cluster and mintmaker namespace), this will raise an error.

Here is an example how we mount a configmap only when it exists: https://github.com/konflux-ci/mintmaker/blob/8dde1c20a51bca9b74dabb42d7defc703b4432bd/pkg/renovate/job.go#L196-L225

qixiang (commented):
Generally, it looks good to me, just a few comments.

I tried to test this, but Renovate is unable to detect updates for an image I used for testing:
```
FROM registry.redhat.io/rhel9/go-toolset@sha256:d128c3d36878251f039606f144ef2608746c3203708b722295e6c571df1d8613
```
it reported: "Could not determine new digest for update (docker package registry.redhat.io/rhel9/go-toolset)"
@staticf0x do you have an example of image which works for testing?

pkg/renovate/job.go
```diff
@@ -79,6 +81,93 @@ func (j *JobCoordinator) getCAConfigMap(ctx context.Context) (*corev1.ConfigMap,
...
+		"auths": mergedAuths,
+	}
+
+	mergedConfigJson, err := json.Marshal(mergedDockerConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	timestamp := time.Now().Unix()
+	name := fmt.Sprintf("merged-registry-image-pull-secret-%d-%s", timestamp, RandomString(5))
```
qixiang, line 151: btw, how about using a shorter name `image-pull-secrets-%d-%s`?

qixiang:
> I tried to test this, but Renovate is unable to detect updates for an image I used for testing:
> ```
> FROM registry.redhat.io/rhel9/go-toolset@sha256:d128c3d36878251f039606f144ef2608746c3203708b722295e6c571df1d8613
> ```
> 
> it reported: "Could not determine new digest for update (docker package registry.redhat.io/rhel9/go-toolset)" @staticf0x do you have an example of image which works for testing?

Please ignore this question. I saw the new jira card you created for extending this, I misunderstood this would also make the dockerfile work with private registries.

gnaponie:
I'm not a go lang expert, but this seems like a good logic and it seems ok by me. And Qixiang's suggestions make sense.

pkg/renovate/job.go
```diff
@@ -79,6 +81,93 @@ func (j *JobCoordinator) getCAConfigMap(ctx context.Context) (*corev1.ConfigMap,
...
+
+	log.Info(fmt.Sprintf("Found %d secrets to merge", len(secretList.Items)))
+
+	mergedAuths := make(map[string]interface{})
+	for _, secret := range secretList.Items {
+		if secret.Type == corev1.SecretTypeDockerConfigJson {
+			data, exists := secret.Data[".dockerconfigjson"]
+			if !exists {
+				// No .dockerconfigjson section
+				continue
```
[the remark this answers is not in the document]
staticf0x: I was thinking that we can attempt to merge as many secrets as we can and skip those that either don't have the correct label (done in filtering before) or don't have the section (which would be incorrect config). Is it still okay to return an error?

pkg/renovate/job.go
```diff
@@ -79,6 +81,93 @@ func (j *JobCoordinator) getCAConfigMap(ctx context.Context) (*corev1.ConfigMap,
...
+			}
+
+			decodedData, err := base64.StdEncoding.DecodeString(string(data))
+			if err != nil {
+				log.Info("Couldn't decode base64 secret data, the secret could be stored in plain text")
+				decodedData = data
+			}
```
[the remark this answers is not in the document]
staticf0x: Thanks for digging into this

pkg/renovate/job.go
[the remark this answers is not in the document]
qixiang: I don't remember clearly, but I have a vague memory, cluster can create some secret automatically for some registries like `127.0.0.1`, so I'd suggest we only merge secrets match the label which is under our control.

If there is no secrets to merge, we can just return `nil, nil`, and skip mounting the volume.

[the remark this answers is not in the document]
qixiang: We can check `mergedAuths` before creating the secret:
```
if len(mergedAuths) == 0 {
        return nil, nil
}
```

FernandesMF:
I have no suggestions from my own (though I'm also a beginner in go).

pkg/renovate/job.go
staticf0x: Fixed, thank you

pkg/renovate/job.go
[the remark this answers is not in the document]
staticf0x: Fixed

staticf0x:
@qixiang I changed the code according to your review, can you PTAL on the new version?

pkg/renovate/job.go
```diff
@@ -224,18 +314,51 @@ func (j *JobCoordinator) Execute(ctx context.Context, tasks []*Task) error {
...
+		job.Spec.Template.Spec.Containers[0].VolumeMounts = append(job.Spec.Template.Spec.Containers[0].VolumeMounts, registrySecretMount)
+	}
+
 	// Create the job
 	if err := j.client.Create(ctx, job); err != nil {
 		return err
 	}
 	log.Info("renovate job created", "jobname", job.Name, "tasks", len(tasks))
+
+	// Set ownership so all resources get deleted once the job finishes
```
qixiang, line 347:
```suggestion
	// Set ownership so all resources get deleted once the job is deleted
```

qixiang (approved):
Good job, just a minor comment to comment :)
The commits should be squashed before merging.

pkg/renovate/job.go
staticf0x: Thank you :) I will squash the commits upon merging.

PR #64, opened 2024-09-11
Transform registry secret to Renovate's hostRules config

staticf0x:
This code uses the previously merged registry secrets (label `mintmaker.appstudio.redhat.com/secret-type: registry`) and parses the JSON to find the `auth` and split it into username and password. Renovate can now successfully make requests to the private registry and create PRs to update Dockerfiles, like
```diff
- FROM registry.redhat.io/ubi9/ubi:9.4-1123
+ FROM registry.redhat.io/ubi9/ubi:9.4-1214.1725849297
```

### DIFF
--- a/pkg/renovate/job.go
+++ b/pkg/renovate/job.go
@@ -181,17 +181,20 @@ func (j *JobCoordinator) Execute(ctx context.Context, tasks []*Task) error {
 	secretTokens := map[string]string{}
 	configMapData := map[string]string{}
 	var renovateCmd []string
+
+	// Create a merged secret for private registries
+	registry_secret, err := j.createMergedPullSecret(ctx)
+
 	for _, task := range tasks {
 		taskId := RandomString(5)
 		secretTokens[taskId] = task.Token
 
-		config, err := task.GetJobConfig(ctx, j.client)
+		config, err := task.GetJobConfig(ctx, j.client, registry_secret)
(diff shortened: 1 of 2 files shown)


pkg/renovate/job.go
```diff
@@ -181,17 +181,20 @@ func (j *JobCoordinator) Execute(ctx context.Context, tasks []*Task) error {
...
 		secretTokens[taskId] = task.Token
 
-		config, err := task.GetJobConfig(ctx, j.client)
+		config, err := task.GetJobConfig(ctx, j.client, registry_secret)
 		if err != nil {
 			return err
 		}
 		configMapData[fmt.Sprintf("%s.json", taskId)] = config
 
-		log.Info(fmt.Sprintf("Creating renovate config map entry with length %d and value %s", len(config), config))
```
staticf0x, line 194: I had to remove this line because the new config now contains unprotected secrets. Is logging this config important? If yes, we would need to first obfuscate the secrets before printing it.

gnaponie: I don't think we care. In case we'll need it for debugging reasons we can put it back obfuscating the secret first.

qixiang: yes, this is not important, we can remove it.

qixiang (approved):
LGTM

PR #67, opened 2024-09-13
Enable rpm-lockfile as allowedPostUpgradeCommands

gnaponie:
Some teams would like to run the rpm-lockfile script as a post-upgrade command. To allow them to do so, we need to add this configuration option.
See CWFHEALTH-3405 for more context on this.

ref: CWFHEALTH-3405

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -180,5 +180,6 @@
     "branchPrefix": "konflux/mintmaker/"
   },
   "forkProcessing": "enabled",
+  "allowedPostUpgradeCommands": ["^rpm-lockfile-prototype rpms.in.yaml$"],
   "dependencyDashboard": false
 }


jpopelka (approved):
Thanks! :1st_place_medal:

ckelleyRH (approved):
LGTM, thanks!

config/renovate/renovate.json
```diff
@@ -180,5 +180,6 @@
     "branchPrefix": "konflux/mintmaker/"
   },
   "forkProcessing": "enabled",
+  "allowedPostUpgradeCommands": ["rpm-lockfile-prototype rpms.in.yaml"],
```
staticf0x, line 183: I just wanted to ask if there is a difference between this and `"^rpm-lockfile-prototype rpms.in.yaml$"` from security point of view? The Renovate docs aren't very clear about this unfortunately.

gnaponie: Updated with the suggestion.

staticf0x (approved):

PR #74, opened 2024-09-19
Fix unit tests

querti:
Due to a recent change, unit tests were failing with a null pointer dereference error. Fix the issue by checking if registry_secret is not nil.


staticf0x (approved):
LGTM

PR #79, opened 2024-09-24
Run MintMaker only on specified components

querti:
Implement an option to only run MintMaker on the components specified in the dependencyupdatecheck CR. Components are specified in a nested structure of workspace->application->component. MintMaker will process all components of a given workspace/application if no lower level entries are specified. The functionality to filter processed components is enabled only if the CR contains any components filters, so the default behavior of MintMaker remains unchanged.

Refers to CLOUDDST-24412

### DIFF
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -17,11 +17,14 @@ limitations under the License.
 package controller
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"strings"
 
 	appstudiov1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
+	mmv1alpha1 "github.com/konflux-ci/mintmaker/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


FernandesMF (approved):
Looks good to me, I have no suggestions to make.
However, I'm far from being a Go expert, it is probably good to wait to hear from the others as well

internal/controller/common.go
```diff
@@ -75,3 +78,50 @@ func getGitProvider(component appstudiov1alpha1.Component) (string, error) {
...
+// Get only components that match a given workspace/application/componentname
+func getFilteredComponents(workspaces []mmv1alpha1.WorkspaceSpec, apiClient client.Client, ctx context.Context) ([]appstudiov1alpha1.Component, error) {
+	components := []appstudiov1alpha1.Component{}
+	err := error(nil)
+
+	// Iterate workspaces and create query filtered by namespace ({workspace}-tenant)
+	for _, workspace := range workspaces {
+		workspaceComponentList := &appstudiov1alpha1.ComponentList{}
+		listOps := &client.ListOptions{
+			Namespace: workspace.Workspace + "-tenant",
```
qixiang, line 91: I'm unsure if this is something hard-coded in Konflux or it's configurable, if this is configurable, this might not work for other konflux clusters (for example the one setup for Fedora community). I think it's fine at this moment, please take a note to recheck this once we have some important features relies on such kind of CRs (for example when we support component nudging in mintmaker).

qixiang (approved):
LGTM, once it's deployed to stage, please create some CRs with workspaces/applications/components to verify this new feature.

PR #80, opened 2024-09-25
Propose updates only to stable Fedora versions

nvtnlucie:
A change to global configuration to not propose Fedora updates beyond stable by replacing updates from https://quay.io/repository/fedora/fedora with https://endoflife.date/fedora .

Example merge request proposed by Mintmaker can be viewed [here](https://gitlab.cee.redhat.com/lnovotna/msaas/-/merge_requests/61).

Refers to CWFHEALTH-3193.


config/renovate/renovate.json
```diff
@@ -78,10 +78,28 @@
       }
     ]
   },
+  "customManagers": [
+    {
+      "customType": "regex",
+      "description": "Fedora updates from a stable source",
+      "fileMatch": ["(^|/|\\.)([Dd]ocker|[Cc]ontainer)file$", "(^|/)([Dd]ocker|[Cc]ontainer)file[^/]*$"],
+      "matchStrings": ["quay.io/fedora/fedora:(?<currentValue>.[0-9]+)"],
```
qixiang, line 86: These fedora images can come from other registries, especially `registry.fedoraproject.org`, and there are different variants, like: `fedora-minimal` and etc.

nvtnlucie: Thats a good point, thank you. I updated the PR to match these cases as well, here is an example of renovate creating [2 separate MRs](https://gitlab.cee.redhat.com/lnovotna/msaas/-/merge_requests) for 2 fedora sources.

config/renovate/renovate.json
```diff
@@ -78,10 +78,28 @@
...
   "dockerfile": {
     "enabled": true,
     "additionalBranchPrefix": "{{baseBranch}}/",
-    "branchPrefix": "konflux/mintmaker/"
+    "branchPrefix": "konflux/mintmaker/",
+    "packageRules": [
+      {
+        "description": "Don't update Fedora to an unstable version",
+        "matchPackagePatterns": ["[\\w\\.\\-\\/]+\\/fedora[\\w\\-]*"],
+        "enabled": false
```
qixiang, line 100: I just realized if users are using tag + digest pinning, it will break after we disabling this, the new regex manager doesn't support it, right?

presets/force-stable-fedora.json
```diff
@@ -0,0 +1,33 @@
+// This preset is used to force Renovate to only suggest updates to stable Fedora releases
```
qixiang, line 1: 1. Could you rename this file with `.json5` suffix?
2. I'd suggest place the new preset under `renovate/presets` directory, or maybe `config/renovate/presets` since some teams are referencing the `config/renovate/renovate.json` , putting them in the same `config` dir seems make more sense, wdyt?

nvtnlucie: Agree with both, thank you. (Not sure why vscode even allowed me the first point :) )

nvtnlucie:
@qixiang @gnaponie @staticf0x I think PR should be good to go now, can you take one last look?

qixiang (approved):
:+1:

gnaponie:
+1 \o/

PR #83, opened 2024-10-09
Document the release process

staticf0x:


querti (commented):
Maybe I would also mention that infra-deployments PRs are created automatically for staging, but manually for prod.

nvtnlucie (approved):

gnaponie:
optional: add some idea on how to test things in stage and check the staging instance to make sure that everything works properly before releasing. 
But it's optional. I believe this also looks fine as it is. +1

staticf0x:
@querti Thanks, added the note.

@gnaponie That sounds like a good thing to write down, however, do we have a unified way how to do this? I have my own personal repo set up in Konflux staging so when I get a new PR from MintMaker I consider the staging to work. Maybe we should talk about having some checklist to verify it?

gnaponie:
The PR looks good, if you could squash the commits it would be even better. But I'm ok with merging this as it is +1

PR #90, opened 2024-10-30
Update config for new version of RPM manager

staticf0x:
Do not merge yet. This requires the new RPM manager updates to be in effect.

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -100,19 +100,21 @@
     "additionalBranchPrefix": "{{baseBranch}}/",
     "branchPrefix": "konflux/mintmaker/"
   },
+  "rpm": {
+    "enabled": true,
+    "packageRules": {
+      "groupName": "RPM updates",
+      "commitMessageAction": "",
+      "commitMessageTopic": "RPM updates",
+      "prTitle": "RPM updates"
+    },
+    "additionalBranchPrefix": "{{baseBranch}}/",
+    "branchPrefix": "konflux/mintmaker/"
+  },
   "git-submodules": {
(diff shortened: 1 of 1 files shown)


gnaponie:
This looks correct.
I see that you've removed a bunch of options about `recreateWhen` and related and also the branchTopic. Is there a specific reason for that?

staticf0x:
I think the diff here is a bit misleading, I removed the whole `lockFileMaintenance` section with all the config options inside and added the `rpm` section. I hope I'm not mistaken, but the `lockFileMaintenance` section was for the previous `rpm` manager only, so we should not need it anymore. It can still be turned on per repo.

qixiang:
The only concern from my side is the source branch name used by the rpm manager, we need to include base branch name in the source branch name of PR/MR. Otherwise, Renovate will use the same branch for proposing changes to different branches, which won't work for many repositories.

staticf0x:
@qixiang I tested the configuration on my repo in gitlab and the branch for rpm-updates is like this: `konflux/mintmaker/main/rpm-updates` so the base branch is already in there and should be safe, WDYT?

I just modified the config a little bit because Renovate had some warnings there.

qixiang:
> @qixiang I tested the configuration on my repo in gitlab and the branch for rpm-updates is like this: `konflux/mintmaker/main/rpm-updates` so the base branch is already in there and should be safe, WDYT?
> 
> I just modified the config a little bit because Renovate had some warnings there.

ah, yes, I didn't notice there are `branchPrefix` and `additionalBranchPrefix` in `rpm` manager config, thank you.

LGTM.

PR #93, opened 2024-11-05
Introduce PipelineRun utils

gnaponie:
This is coming from the konflux-ci/release-service repo. Originally this is part of a shared set of utils.
These methods should be useful in the future to build the PipelineRun definition.

ref: CWFHEALTH-3293

### DIFF
--- a/pkg/tekton/utils.go
+++ b/pkg/tekton/utils.go
@@ -1,0 +1,33 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tekton
+
+import (
+	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	"knative.dev/pkg/apis"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// hasPipelineSucceeded returns a boolean indicating whether the PipelineRun succeeded or not.
+// If the object passed to this function is not a PipelineRun, the function will return false.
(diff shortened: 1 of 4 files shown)


gnaponie:
@qixiang hey, could you check this? I've marked this as a draft, as I'm not sure this is complete. Thanks.

qixiang (commented):
Since this new package is not called by any code yet, I think it's fine to add them as a new package.

The new package can be moved to the `pkg` directory. However, we should note that:

- Our current packages in `pkg` are not properly designed for external usage - they contain APIs that are tightly coupled with the mintmaker project.
- Due to this coupling, we'll probably move the `pkg` to `internal/pkg` to better reflect their non-reusable nature.

Another thing is we should create a new branch in this project (e.g. `refactoring-pipelinerun`), then open the PR against the new branch.

gnaponie:
@qixiang thanks. I've addressed all the comments.

gnaponie:
@qixiang thanks. I've addressed the comments in the files.
About your other comment, just to make it clear before I make the change:
* We should add `internal/pkg` 
* `internal/pkg` directory should include the `tekton` pkg/directory. Or should that remain in `pkg` (main directory) since it's independent by MintMaker? 
* `controller` (now inside `internal`) should be moved to `internal/pkg`
* should we move all pkgs inside `pkg` to `internal/pkg` ? perhaps something should remain there (thinking about `k8s` as an example).


About the branch, yes, I'll close this one and open it against a different branch.

qixiang:
Hi @gnaponie,

We can add `tekton` in `pkg`, forget about moving `pkg` to `internal/pkg`, I just want to remind we'll do that in the future, because the current packages in `pkg` are not implemented to be reusable for other projects. I believe the `tekton` will also have some code tightly coupled with mintmaker, thought it currently doesn't.

> controller (now inside internal) should be moved to internal/pkg
`controller` should remains in `internal`, we don't need to change it.

So about the code structure, the only change is adding `tekton` in `pkg`, instead of having a new top level `tekton` dir.

gnaponie:
@qixiang hey, I've addressed the suggestions. I've also changed the target branch, so we don't mess around with main.
Also this isn't marked as draft anymore. So it is ready to be merged in case this looks fine now.

qixiang:
@gnaponie the new branch `refactoring-pipelinerun` was not created from `main` branch, it includes your changes in this PR. Beside of that:
1. go.mod should be updated to include the new dependencies
2. the tests are failing, we don't have tests enabled in this repo yet, so you need to run them locally by manual.

PR #95, opened 2024-11-06
Fix platformCommit value

staticf0x:
It is now a string: https://docs.renovatebot.com/configuration-options/#platformcommit


qixiang (approved):
LGTM

PR #97, opened 2024-11-14
Reintroduce lockFileMaintenance section for RPM manager

staticf0x:
This was needed after all


qixiang:
Could you elaborate on why this is required?

staticf0x:
If you only have `rpms.in.yaml` and no lock file yet, the only way to create it is through lockFileMaintenance. However individual updates also require this. The MR table with updates will always contain "All locks refreshed" and then individual RPM updates.

qixiang (approved):
Thanks for the explanation, LGTM.

PR #98, opened 2024-11-15
Enable Go managers

nvtnlucie:
2 Go managers are added in this commit. I also moved the `branchPrefix` attribute one level higher as it seems from the documentation the effect stays the same.

Closes [CWFHEALTH-3173](https://issues.redhat.com/browse/CWFHEALTH-3173)
Related PR: https://github.com/konflux-ci/mintmaker-renovate-image/pull/71

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -313,11 +313,17 @@
   },
   "gomod": {
     "additionalBranchPrefix": "{{baseBranch}}/",
-    "branchPrefix": "konflux/mintmaker/"
+    "branchPrefix": "konflux/mintmaker/",
+    "schedule": [
+      "after 5am on sunday"
+    ]
   },
   "ocb": {
     "additionalBranchPrefix": "{{baseBranch}}/",
-    "branchPrefix": "konflux/mintmaker/"
+    "branchPrefix": "konflux/mintmaker/",
+    "schedule": [
+      "after 5am on sunday"
+    ]
   },
   "forkProcessing": "enabled",
   "allowedPostUpgradeCommands": [
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -321,8 +321,6 @@
     ]
   },
   "gomod": {
-    "additionalBranchPrefix": "{{baseBranch}}/",
-    "branchPrefix": "konflux/mintmaker/",
     "schedule": [
       "after 5am on sunday"
     ],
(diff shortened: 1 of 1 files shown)


config/renovate/renovate.json
```diff
@@ -109,153 +109,19 @@
...
-    "additionalBranchPrefix": "{{baseBranch}}/",
-    "branchPrefix": "konflux/mintmaker/"
-  },
-  "setup-cfg": {
-    "additionalBranchPrefix": "{{baseBranch}}/",
-    "branchPrefix": "konflux/mintmaker/"
+    "enabled": true
   },
+  "additionalBranchPrefix": "{{baseBranch}}/",
+  "branchPrefix": "konflux/mintmaker/",
```
querti, line 124: Does this change have an effect on tekton manager branches?

gnaponie:
Perhaps it's just github diff changes that is a bit chaotic. But it seems like `homebrew` used to be explicitly enabled and not it isn't anymore. Any reason for removing that or is this a mistake?

qixiang:
I'd suggest splitting the `additionalBranchPrefix` + `branchPrefix` change into a separate PR, we can probably merge it first after checking the comments from Giulia and Lubomir.

We should test the go manager more carefully as this manager is a little complicated and users opened several requests regarding this manager (though they were opened against the other internal renovate instance).

staticf0x:
`tekton` manager needs this line added:
```json
"additionalBranchPrefix": "",
```

because of this error:
```
E       AssertionError: assert 'konflux/references/main/main' == 'konflux/references/main'
E         
E         - konflux/references/main
E         + konflux/references/main/main
E         ?                        +++++
```

staticf0x:
Btw I think it's fine to keep all changes in this PR, but it really helps in this case to change GitHub's diff view to "split" (= side by side).

nvtnlucie:
Thank you for the review, i split the change into 2 PRs, rebased and added schedules as well. can you take another look please?

staticf0x:
I tested `gomod` in three situations:
- Simple repo, 3 deps, current Go version
- Simple repo, 3 deps, older Go version
- Real repo (https://github.com/arcalot/arcaflow-engine-deployer-podman)

and all worked, but it needs the following config to work for all cases:

```json
"gomod":
    {
        "postUpdateOptions":
        [
            "gomodUpdateImportPaths",
            "gomodTidy"
        ]
    },
    "packageRules":
    [
        {
            "matchManagers":
            [
                "gomod"
            ],
            "matchDepTypes":
            [
                "indirect"
            ],
            "enabled": true
        }
    ]
```

staticf0x (approved):
Looks good, tests passed, I think we can merge it if others agree.

nvtnlucie:
Converted to draft until optimizations are in place so we dont exceed renovate's runtime.

PR #101, opened 2024-11-19
Use pipelinerun instead of jobs

gnaponie:
ref: CWFHEALTH-3293


internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -18,55 +18,216 @@ package controller
...
+
+	// Create a merged secret for private registries
+	registry_secret, err := r.createMergedPullSecret(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var renovateCmd []string
+	// TODO: this needs to be changed to define the RENOVATE_TOKEN once we will get the secret for the component
+	for _, task := range tasks {
```
qixiang, line 166: `tasks` is not defined. To keep things simple at this stage, I suggest we avoid moving the code from `renovate/job.go` to the new controller for now. Instead, focus on making the new controller(s) create a simple `PipelineRun` that only prints a message, as you're currently doing.

There are other undefined variables in this PR as well.

After preparing the new controller(s), you can try to run it locally with (if you have a local `minikube` cluster):
```
$ kubectl apply -k https://github.com/konflux-ci/application-api/config/crd

$ kubectl apply -k https://github.com/redhat-appstudio/infra-deployments/components/mintmaker/development

# then let's remove the old controller by removing the deployment
$ kubectl delete deployment/mintmaker-controller-manager

# now we can run the new controller(s) by manual
$ go run cmd/main.go
```

gnaponie: Thank you for the review. I agree, I also tried to run this locally, but still it doesn't run, I still have some errors and tests aren't defined. That's how far I've got on Friday. I'll follow your suggestions and perhaps we can merge it, and then I'll have another PR to continue with this work. But at least if other folks wants to contribute they aren't blocked anymore.

gnaponie:
Hey @qixiang, I've tested this on my minikube cluster. I confirm you that this is running correctly. I was able to create a dependencyupdatecheck CR and a new pending pipelinerun was created. I'm working on the tests and then this PR is ready to a final review and to be merged.

gnaponie:
@qixiang @FernandesMF @staticf0x @querti @nvtnlucie I think this is now ready for a review

internal/controller/pipelinerun_controller_test.go
```diff
@@ -0,0 +1,32 @@
...
+limitations under the License.
+*/
+
+package controller
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("Pod Controller", func() {
```
qixiang, line 23:
```suggestion
var _ = Describe("PipelineRun Controller", func() {
```

pkg/renovate/task.go
```diff
@@ -117,7 +117,7 @@ func (t *Task) TransformHostRules(ctx context.Context, registrySecret *corev1.Se
 	return hostRules, nil
 }
 
-func (t *Task) GetJobConfig(ctx context.Context, client client.Client, registrySecret *corev1.Secret) (string, error) {
+func (t *Task) GetPipelineRunConfig(ctx context.Context, client client.Client, registrySecret *corev1.Secret) (string, error) {
```
qixiang, line 120: This is not used at all at this moment, right? then we don't need to include it.

Makefile
```diff
@@ -212,7 +212,7 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.2.1
-CONTROLLER_TOOLS_VERSION ?= v0.13.0
+CONTROLLER_TOOLS_VERSION ?= v0.14.0
```
qixiang, line 215: I'd suggest have a separate commit for the controller-tools version changes (Makefile, README.md, config/crd/bases/appstudio.redhat.com_dependencyupdatechecks.yaml), open a PR to main branch instead.

README.md
```diff
@@ -72,6 +72,29 @@ make uninstall
...
+```
+
+and in the trace you can see that the `controller-tools` version `v0.13.0` is being used, make sure to upgrade to `v0.14.0`.
+To do that, run:
+
+```sh
+go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0
+GOBIN=$PWD/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0
+./bin/controller-gen --version
+```
```
qixiang, line 96:
```suggestion
```
By checking the Makefile, when the version doesn't match your updated `v0.14.0`, `make test` should be able to install the correct version automatically. So I'd suggest to remove this paragraph.

Makefile
gnaponie: see https://github.com/konflux-ci/mintmaker/pull/113

gnaponie:
All comments addressed.

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -19,54 +19,95 @@ package controller
...
+				},
+			},
+		},
+	}
+
+	if err := r.Client.Create(ctx, pipelineRun); err != nil {
+		return nil, err
+	}
+
+	log.Info(fmt.Sprintf("Created pipelinerun %s", name))
```
qixiang, line 100:
```suggestion
```
Remove this to avoid dup log lines, we also print the log in reconcile.

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -96,99 +137,32 @@ func (r *DependencyUpdateCheckReconciler) Reconcile(ctx context.Context, req ctr
...
-		}
-	}
-
-	if len(tasks) == 0 {
-		return ctrl.Result{}, nil
-	}
-
-	log.Info("executing renovate tasks", "tasks", len(tasks))
-	err = r.jobCoordinator.ExecuteWithLimits(ctx, tasks)
+	log.Info("Creating pending pipeline runs")
```
qixiang, line 146:
```suggestion
	log.Info("Creating pending PipelineRun")
```

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -96,99 +137,32 @@ func (r *DependencyUpdateCheckReconciler) Reconcile(ctx context.Context, req ctr
...
+	pipelinerun, err := r.createPipelineRun(ctx)
 	if err != nil {
-		log.Error(err, "failed to create a job")
+		log.Error(err, "failed to create pipelineruns")
```
qixiang, line 149:
```suggestion
		log.Error(err, "failed to create PipelineRun")
```

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -96,99 +137,32 @@ func (r *DependencyUpdateCheckReconciler) Reconcile(ctx context.Context, req ctr
...
+	} else {
+		log.Info(fmt.Sprintf("Created pipelinerun %v", pipelinerun))
```
qixiang, line 151:
```suggestion
		log.Info(fmt.Sprintf("Created PipelineRun %s", pipelinerun.Name))
```

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -96,99 +137,32 @@ func (r *DependencyUpdateCheckReconciler) Reconcile(ctx context.Context, req ctr
 	}
 	dependencyupdatecheck.Annotations[MintMakerProcessedAnnotationName] = "true"
 
-	err = r.client.Update(ctx, dependencyupdatecheck)
+	err = r.Client.Update(ctx, dependencyupdatecheck)
```
qixiang, line 140:
```suggestion
        patch := client.MergeFrom(dependencyupdatecheck.DeepCopy())
	err = r.Client.Patch(ctx, dependencyupdatecheck, patch)
```
I found the update doesn't work in my env, it reports error of:
```
Operation cannot be fulfilled on dependencyupdatechecks.appstudio.redhat.com \"dependencyupdatecheck-svjrb\": the object has been modified; please apply your changes to the latest version and try again
```
but I haven't found the root cause, but changing it to `Patch` works.

gnaponie:
Comments addressed.

PR #103, opened 2024-11-20
Update schedules for the managers

gnaponie:
There isn't a reason to run some managers every day
multiple time per day. So let's run some of them only once per week.
Let's just run the rpm manager everyday, since this might contain
some CVE fix we want to address during the day.
Also, let's pick different days for different managers,
this is hacky, but it's a good enough and very simple solution
to "randomize" managers executions before we have the modularity
epic in place.

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -93,12 +93,18 @@
         "recreateWhen": "always",
         "rebaseWhen": "behind-base-branch"
       }
+    ],
+    "schedule": [
+      "after 5am on saturday"
     ]
   },
   "dockerfile": {
     "enabled": true,
     "additionalBranchPrefix": "{{baseBranch}}/",
-    "branchPrefix": "konflux/mintmaker/"
+    "branchPrefix": "konflux/mintmaker/",
+    "schedule": [
+      "after 5am on monday"
+    ]
   },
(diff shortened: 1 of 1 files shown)


gnaponie:
@qixiang @querti @staticf0x @FernandesMF @nvtnlucie PTAL

config/renovate/renovate.json
```diff
@@ -93,6 +93,9 @@
         "recreateWhen": "always",
         "rebaseWhen": "behind-base-branch"
       }
+    ],
+    "schedule": [
+      "after 5am on saturday"
```
qixiang, line 98: Do you know if we can apply `updateNotScheduled` to tekton manager only? Updates to existing PRs can also trigger PLRs, which is probably not what users want.

staticf0x: Yes it can be used inside `packageRules` to tekton only

gnaponie: I've rebase the PR. Would it be useful in all the managers?

gnaponie:
I've updated the PR. Please refer to the updated PR description to get more insights on this.

staticf0x:
I think this is good for now, LGTM

staticf0x (approved):

PR #105, opened 2024-11-26
Update tekton manager schedule

qixiang:
Renovate from MintMaker updated its global schedules. The Tekton manager only runs on Saturdays, so there's no need to specify a custom schedule. Defining a schedule outside the global schedule window could also prevent pull requests from being merged.


gnaponie:
+1

PR #108, opened 2024-11-26
Enable pep723 manager

staticf0x:
This requires renovate >=38.104.0, once we are there, we can enable this manager 
cc @FernandesMF


staticf0x:
We are now on 38.132.0, so we can merge this

gnaponie (approved):

PR #114, opened 2024-12-06
Pipelinerun controller (draft, don't merge!)

FernandesMF:
Hi team, I finished a first draft of the new PipelineRun controller.

This controller has the responsibility of checking how many pipelineruns are running at the moment, and if there are fewer than a specified value, kick new pipelineruns until the maximum is reached.

Could you please review the code and give your suggestions? Thanks! 😁


qixiang (commented):
Just minor comments.
Could you rebase it to latest `refactoring-pipelinerun` branch? which could make it much easier for reviewing.

internal/controller/pipelinerun_controller.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *PipelineRunReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&tektonv1beta1.PipelineRun{}).
+		/* TODO: For the time being we just ignore all types of events
```
gnaponie, line 103: this comment can be removed.

internal/controller/pipelinerun_controller_test.go
```diff
@@ -0,0 +1,32 @@
...
+package controller
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("Pod Controller", func() {
+	Context("When reconciling a resource", func() {
+
+		It("should successfully reconcile the resource", func() {
```
gnaponie, line 26: We can keep this for a second time, but let's also write some tests. I think it will be sufficient to test that when a PipelineRun finishes a new one starts.

pipelinerun.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: tekton.dev/v1beta1
```
gnaponie, line 1: Probably this file doesn't need to be pushed. If needed we can create a directory such as `utils` (maybe with a better name) and we can include some useful files there.

internal/controller/pipelinerun_controller.go
```diff
@@ -0,0 +1,114 @@
...
+func (r *PipelineRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrllog.FromContext(ctx).WithName("PipelineRun")
+	ctx = ctrllog.IntoContext(ctx, log)
+
+	if req.Namespace != MintMakerNamespaceName {
+		return ctrl.Result{}, nil
+	}
+
+	// Get pipelineruns
+	var childPipelineRuns tektonv1beta1.PipelineRunList
```
gnaponie, line 62: Is `childPipelineRuns` a list of all the PipelineRuns (in whatever state)? I'm not sure if `child` is an appropriate name. Perhaps let's choose a more meaningful? Maybe `existingPipelineRuns`, I'm actually terrible at names... but `child` makes me think that there's a parent.

gnaponie:
I've also added only minor comments, the change overall looks good! I've checked only the pipelinerun controller, as I'm assuming the rest is from previous commits. But as Qixiang said, let's rebase the branch, also because my changes were merged, so now they are more correct.

FernandesMF:
@gnaponie @qixiang thank you for your reviews, I really appreciate them!
I'll address all the points you raised. In the meantime, I have rebased on top of `refactor-pipeline`, feel free to have another look if you want. (And sorry for not having rebased before... I just drafted the MR and didn't realise it was a almost 1k lines long 😅 )

internal/controller/pipelinerun_controller.go
FernandesMF: Good call! I like `existing` better

qixiang (commented):
Could you rebase it on top of the latest "refactoring-pipelinerun` branch, so we can focus on the necessary changes?

qixiang:
@FernandesMF can we close this PR?

PR #120, opened 2024-12-10
Pin go-toolset image to version 1.21

qixiang:


.github/renovate.json
```diff
@@ -4,5 +4,15 @@
     "automergeType": "pr",
     "automerge": true,
     "platformAutomerge": true
+  },
+  "dockerfile": {
+    "packageRules": [
+      {
+        "packageNames": [
```
staticf0x, line 11: Wait, shouldn't it be `matchPackageNames`?

staticf0x (changes requested):
See my above comment

.github/renovate.json
qixiang: You're correct, `packageNames` is the old name, we should avoid using that, fixed.

staticf0x (approved):

PR #121, opened 2024-12-10
Fix git URL parsing

nvtnlucie:
Fixes a bug where the controller crashed if git url was provided in format `git@...` by replacing the 2 occurences of `url.Parse()` with more sophisticated function. The error causing controller to crash is happening inside `NewScmComponent()`. The code inside `getGitProvider()` did not produce an error, however it always jumped into the second block, relying on component annotations.

Closes [CWFHEALTH-3811](https://issues.redhat.com/browse/CWFHEALTH-3811).

### DIFF
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -19,11 +19,11 @@ package controller
 import (
 	"context"
 	"fmt"
-	"net/url"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


qixiang (approved):
LGTM

PR #123, opened 2024-12-12
Introduce PipelineRun builder - tekton pkg

gnaponie:
Introduce a set of utils to build PipelineRuns.
This will be added as an internal pkg called `tekton`.


gnaponie:
@qixiang @FernandesMF PTAL

qixiang (approved):

PR #124, opened 2024-12-12
Add internal component package

qixiang:
Add new internal component package that provides Go structs to encapsulate Konflux components and APIs for retrieving component information.

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -21,8 +21,10 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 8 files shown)


qixiang:
This is missing docs and tests, but we want to have this skeleton code to work other features on top of, I'll add docs and tests later. Once we have the feature ready to be merged to main branch, we'll squash commits in the `refactoring-pipelinerun` branch, so such ugly code won't be in main branch :)

gnaponie (approved):

gnaponie:
As we discussed: even if there's room for improvement, let's merge this as it is, so we can continue working in parallel on this. As you said, we'll squash it later.

PR #130, opened 2024-12-16
Mount self-signed certificate

gnaponie:
For some hosts, we need to mount self-signed certificates to run Renovate and other dependent tools. Let's mount this ConfigMap to in all PipelineRun.

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -57,6 +57,29 @@ func NewDependencyUpdateCheckReconciler(client client.Client, scheme *runtime.Sc
 	}
 }
 
+// getCAConfigMap returns the first ConfigMap found in mintmaker namespace
+// that has the label 'config.openshift.io/inject-trusted-cabundle: "true"'.
+// If no such ConfigMap is found, it returns nil.
+func (r *DependencyUpdateCheckReconciler) getCAConfigMap(ctx context.Context) (*corev1.ConfigMap, error) {
+	configMapList := &corev1.ConfigMapList{}
+	labelSelector := client.MatchingLabels{"config.openshift.io/inject-trusted-cabundle": "true"}
+	listOptions := []client.ListOption{
+		client.InNamespace(MintMakerNamespaceName),
+		labelSelector,
+	}
+	err := r.Client.List(ctx, configMapList, listOptions...)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(configMapList.Items) > 0 {
+		// Just return the configmap
+		return &configMapList.Items[0], nil
+	}
+
+	return nil, nil
+}
+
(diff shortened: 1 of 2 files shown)


gnaponie:
@qixiang @FernandesMF PTAL

FernandesMF (approved):

FernandesMF:
Looks good to me!

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -126,6 +150,21 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(comp component.GitCo
 	secretOpts := utils.NewMountOptions().WithTaskName("build").WithStepNames([]string{"renovate"})
 	builder.WithSecret(name, "/etc/renovate/secret", secretItems, secretOpts)
 
+	// Check if a ConfigMap with the label `config.openshift.io/inject-trusted-cabundle: "true"` exists.
+	// If such a ConfigMap is found, add a volume to the PipelineRun specification to mount this ConfigMap.
+	// The volume will be mounted at '/etc/pki/ca-trust/extracted/pem' within the PipelineRun Pod.
+	caConfigMap, _ := r.getCAConfigMap(ctx)
```
nvtnlucie, line 156: Do we want to ignore the error here? If yes, maybe the method does not need to return err at all.

nvtnlucie (commented):
LGTM as well except for one suggestion

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -57,6 +57,29 @@ func NewDependencyUpdateCheckReconciler(client client.Client, scheme *runtime.Sc
...
+// that has the label 'config.openshift.io/inject-trusted-cabundle: "true"'.
+// If no such ConfigMap is found, it returns nil.
+func (r *DependencyUpdateCheckReconciler) getCAConfigMap(ctx context.Context) (*corev1.ConfigMap, error) {
+	configMapList := &corev1.ConfigMapList{}
+	labelSelector := client.MatchingLabels{"config.openshift.io/inject-trusted-cabundle": "true"}
+	listOptions := []client.ListOption{
+		client.InNamespace(MintMakerNamespaceName),
+		labelSelector,
+	}
+	err := r.client.List(ctx, configMapList, listOptions...)
```
qixiang, line 70:
```suggestion
	err := r.Client.List(ctx, configMapList, listOptions...)
```

internal/controller/dependencyupdatecheck_controller.go
gnaponie: If we catch an error in this phase, we don't need to exit/return from the function, since we can just disregard it. But perhaps some logs might be helpful for debugging reasons. I'll add that.

gnaponie:
All comments addressed.

qixiang (approved):
LGTM

PR #131, opened 2024-12-17
Include component specific info in Renovate config

gnaponie:


gnaponie:
@qixiang @FernandesMF @nvtnlucie PTAL

internal/pkg/component/gitlab/gitlab.go
```diff
@@ -165,3 +172,34 @@ func (c *Component) getDefaultBranch() (string, error) {
...
+	baseConfig, err := c.GetRenovateBaseConfig(c.client, c.ctx)
+	if err != nil {
+		return "", err
+	}
+
+	baseConfig["platform"] = c.Platform
+	baseConfig["endpoint"] = c.GetAPIEndpoint()
+	// We don't need to set a username or gitAuthor for gitlab, since this is tight to a token
+	baseConfig["username"] = ""
+	baseConfig["gitAuthor"] = ""
```
qixiang, line 186: I'm a little unsure about `gitAuthor` for gitlab repositories, maybe it's required, but we can update later after testing.

FernandesMF (approved):
Makes sense and looks good to me!

qixiang (approved):
LGTM

internal/pkg/component/gitlab/gitlab.go
gnaponie: Agreed, let's check during the testing phase. I'll add a note in the card. And merge this PR.
Thank you both for the review.

PR #132, opened 2024-12-17
Implement new PipelineRuns controller

FernandesMF:
This commit implements a new controller/reconciler for PipelineRuns. Its main responsibility is to launch new pipelineruns when another one finishes, and try to keep up to a maximum number of pipelineruns active at any given time.

Besides that, it also launches github pipelineruns first, as a temporary means of decreasing the chance that their temporary token will expire.

### Testing remarks
The new controller worked fine in my manual tests. I created several pipelineruns, and watched how they were progressively launched.

The unit tests are having an intermittent behavior. About half of the times I ran them, they fail during the `BeforeEach` phases, seemingly because they fail to create the pipelineruns (or maybe to retrieve them with `listPipelineRuns`). I guess we will need to debug this function to solve this, so maybe we can leave it to another task in the epic.

### DIFF
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -22,7 +22,9 @@ import (
 
 	. "github.com/konflux-ci/mintmaker/pkg/common"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"knative.dev/pkg/apis"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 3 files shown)


internal/controller/pipelinerun_controller_test.go
```diff
@@ -14,19 +14,138 @@ See the License for the specific language governing permissions and
...
+		_ = BeforeEach(func() {
+
+			MaxSimultaneousPipelineRuns = 2
+
+			for i := range 3 {
+				pplrName := "pplnr" + strconv.Itoa(i)
+				pipelineRunBuilder := local_tekton.NewPipelineRunBuilder(pplrName, MintMakerNamespaceName)
+				pipelinerun, err := pipelineRunBuilder.Build()
+				Expect(err).NotTo(HaveOccurred())
+				k8sClient.Create(ctx, pipelinerun)
```
qixiang, line 51: You should check this call result:
```
Expect(k8sClient.Create(ctx, pipelinerun)).Should(Succeed())
```
with this, you will see the pipelineruns are not created successfully, due to "mintmaker" namespace not exists, so you should create the namespace first:
```
createNamespace(MintMakerNamespaceName)
```

internal/controller/pipelinerun_controller_test.go
```diff
@@ -14,19 +14,138 @@ See the License for the specific language governing permissions and
...
+		It("should successfully launch new pipelineruns", func() {
+
+			reconciler := PipelineRunReconciler{
+				Client: k8sClient,
+				Scheme: scheme.Scheme,
+			}
+			reconciler.SetupWithManager(k8sManager)
+			reconciler.Reconcile(ctx, ctrl.Request{
+				NamespacedName: types.NamespacedName{Namespace: MintMakerNamespaceName, Name: "pplnr1"},
+			})
```
qixiang, line 75: As mentioned, you should start the controller when the test env is started, instead of calling the reconcile in tests.
You should start the controller similar to this:
https://github.com/konflux-ci/mintmaker/blob/refactoring-pipelinerun/internal/controller/suite_test.go#L101

internal/controller/pipelinerun_controller.go
```diff
@@ -19,32 +19,99 @@ package controller
...
+	ctx = ctrllog.IntoContext(ctx, log)
+
+	if numToLaunch <= 0 {
+		return nil
+	}
+	numLaunched := 0
+	for _, pipelineRun := range existingPipelineRuns.Items {
+		if pipelineRun.IsPending() {
+			pipelineRun.Spec.Status = ""
+			err := r.Client.Update(ctx, &pipelineRun, &client.UpdateOptions{})
```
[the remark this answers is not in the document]
FernandesMF: Awesome idea! These errors popped up everyone in a while indeed, thanks for this suggestion! 😁 
I'll definitely be more careful with `Update` from now on 😂

internal/controller/pipelinerun_controller_test.go
FernandesMF: Good catch! This fixed the tests indeed, thanks a lot!! 🥳

internal/controller/pipelinerun_controller_test.go
FernandesMF: Great catch as well! I was not checking the calls here and was missing more errors.
Thanks for finding this, now everything works great!

FernandesMF:
Thank you so much for the suggestions, @qixiang! Now the tests are working fine.
I allowed for up to 3 retries, in case of failures. Sometimes (about 1 in 10) the Reconciler seems to be a bit too fast or slow and an unexpected number of pipelineruns are active.

Could you take another look? If there are no further suggestions, I guess this is ready to be merged!

internal/controller/pipelinerun_controller_test.go
```diff
@@ -14,19 +14,119 @@ See the License for the specific language governing permissions and
...
+			var checkPipelineRun tektonv1.PipelineRun
+			for _, p := range existingPipelineRuns.Items {
+				if p.ObjectMeta.Name == "pplnr1" {
+					checkPipelineRun = p
+					break
+				}
+			}
+			// To count running pipelineruns, we have to check for empty 'status' fields;
+			// During the tests, these pipelineruns will not be updated to be actually 'running'
+			Expect(string(checkPipelineRun.Spec.Status)).To(Equal(""))
```
qixiang, line 129: This doesn't work for the purpose, you should check all github PipelineRuns should have start time later than non-github PipelineRuns. 

And actually if you check the PipelineRuns, you will see all have been started by the controller, for example with this test, it will fail:
```
pipelineRuns := listPipelineRuns(MintMakerNamespaceName)
for _, plr := range pipelineRuns {
        if plr.Labels[MintMakerAppstudioLabel] == "github" {
                Expect(plr.Spec.Status).To(BeEmpty())
        } else {
                Expect(plr.Spec.Status).To(Equal(tektonv1.PipelineRunSpecStatusPending))
        }
}
```

FernandesMF: Good catch! This testing logic is much more complete, thanks! I updated this test case with your suggestion.

When I ran it, it failed but it was due to a type mismatch. After I cast `plr.Spec.Status` to string, it worked.
I also included two boolean flags, to make sure both clauses were executed (and thus the non-github pipeline is indeed pending):
https://github.com/konflux-ci/mintmaker/blob/pipelinerun-controller-2/internal/controller/pipelinerun_controller_test.go#L115-L127

Do you think the new test is complete enough? And is is passing correctly for you as well?

qixiang: This still doesn't work for me because both the controller removed pending status from both github and non-github PipelineRuns. I believe the reason is in our testenv, there is actually no Tekton operator running for moving the PipelineRuns to "Running" status once the `PipelineRunSpecStatusPending` status is removed. So our controller always think there is no PipelineRUns are running there, thus it will "start" all PipelineRuns.

There are two possible solutions:
1. Implement a fake Tekton operator to start PipelineRuns (change the related fields `Status.Conditions.*` and etc)
2. Change the corresponding fields in PipelineRuns in our tests, for example:
```
BeforeEach(func() {
...
# max running num is 20
# create 20 PipelineRuns which are in "running" status.  Simulate the pipeline in a running state by setting the corresponding `Status.Conditions.*`  fields. 
# create 10 github PipelineRuns with "pending" status
# create 10 non-github PipelineRuns with "pending" status

It("should start github PipelineRuns first", func() {
...
# update 10 "running" PipelineRun to "completed" status.
# verify there are 20 PipelineRuns in "running" status again, and include 10 github PipelineRuns, all the 10 non-github PipelineRuns remain in "pending" status.
```
I believe the second solution is simpler and provides more flexibility for writing our tests.

qixiang: I'm fine with merging this PR now and address the test improvements later. If you proceed with merging, please add TODO/FIXME comments in the test file to indicate the needed improvements.

gnaponie: +1 to merge this as it is. Also the other tests are all broken or incomplete. Let's just fix them all together and let's work on higher priority cards.

FernandesMF: Okay, I left a comment in that test case, and I'll proceed with merging this change as is. Thanks a lot for the input!
https://github.com/konflux-ci/mintmaker/pull/132/files#diff-450c910037cdd118f7ace46e459a73ccf20c3a163215ed441627c1cd5e51e2adR112-R113

PR #133, opened 2024-12-19
Manage registry access credentials in PipelineRuns

gnaponie:
Secure registries require credentials for image pulling, currently we create the secret in mintmaker namespace and use that for all components, we'll follow the same approach in this refactoring, but modify it to mount the secret in PipelineRuns.

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -34,6 +35,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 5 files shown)


gnaponie:
@qixiang @FernandesMF @nvtnlucie PTAL

FernandesMF (approved):
Looks good to me, I have no suggestion!

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -80,14 +82,103 @@ func (r *DependencyUpdateCheckReconciler) getCAConfigMap(ctx context.Context) (*
...
+	return newSecret, nil
+}
+
 // createPipelineRun creates and returns a new PipelineRun
 func (r *DependencyUpdateCheckReconciler) createPipelineRun(comp component.GitComponent, ctx context.Context) (*tektonv1.PipelineRun, error) {
 
 	log := ctrllog.FromContext(ctx).WithName("DependencyUpdateCheckController")
 	ctx = ctrllog.IntoContext(ctx, log)
 	name := fmt.Sprintf("renovate-%d-%s-%s", comp.GetTimestamp(), RandomString(8), comp.GetName())
+	registry_secret, err := r.createMergedPullSecret(ctx)
```
qixiang, line 179: Currently, we obtain registry credentials from secrets prepared in mintmaker, and these credentials remain same throughout the controller's lifecycle. Since the credentials don't change, it's unnecessary to create a new secret for each PipelineRun. Instead, we can create a single secret and reuse it. For example, we can create the secret during the first reconcile, cache it, and avoid recreating it in subsequent reconciles. This cached secret can then be mounted to all PipelineRuns.

It was not a big issue we create one secret for each job, because we only have 1-5 jobs in each round of Renovate run, but now we have thousands of PipelineRuns in each round.

nvtnlucie (commented):
LGTM as well.

internal/controller/dependencyupdatecheck_controller.go
gnaponie: I had this doubt. Thanks for raising the concern. Let me fix that.

gnaponie:
@qixiang I'm pretty sure it's wrong :) But I've moved the github cache pkg in the base pkg, so that it can be used by both github and the controller. And I've used the cache. Since I'm new to go and the sync pkg I'm sure this is wrong. But it makes somehow sense and the compile doesn't complain :)

gnaponie:
@qixiang as discussed I've rolled back to [fc1ca91](https://github.com/konflux-ci/mintmaker/commit/fc1ca916f9b2c0eed624e9fa1fdfde61537dca0f) -- I think we should we ok now. PTAL

qixiang (approved):
LGTM

PR #135, opened 2025-01-09
Configure pipeline-migration-tool run for tekton manager

tkdchen:
STONEBLD-2824

Executable script pipeline-migration-tool is available from the Mintmaker image.

This change works with https://github.com/konflux-ci/mintmaker-renovate-image/pull/85 together.

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -116,7 +116,13 @@
     ],
     "schedule": [
       "after 5am on saturday"
-    ]
+    ],
+    "postUpgradeTasks": {
+      "commands": [
+        "pipeline-migration-tool -u '[{{#each upgrades}}{\"depName\": \"{{{depName}}}\", \"currentValue\": \"{{{currentValue}}}\", \"currentDigest\": \"{{{currentDigest}}}\", \"newValue\": \"{{{newValue}}}\", \"newDigest\": \"{{{newDigest}}}\", \"packageFile\": \"{{{packageFile}}}\", \"parentDir\": \"{{{parentDir}}}\", \"depTypes\": [{{#each depTypes}}\"{{{this}}}\"{{#unless @last}},{{\/unless}}{{\/each}}]}{{#unless @last}},{{\/unless}}{{\/each}}]'"
+      ],
+      "executionMode": "branch"
+    }
   },
   "dockerfile": {
     "enabled": true,
@@ -352,7 +358,8 @@
   },
   "forkProcessing": "enabled",
   "allowedCommands": [
-    "^rpm-lockfile-prototype rpms.in.yaml$"
+    "^rpm-lockfile-prototype rpms.in.yaml$",
+    "^pipeline-migration-tool -u '\\[{{#each upgrades}}{\"depName\": \"{{{depName}}}\", \"currentValue\": \"{{{currentValue}}}\", \"currentDigest\": \"{{{currentDigest}}}\", \"newValue\": \"{{{newValue}}}\", \"newDigest\": \"{{{newDigest}}}\", \"packageFile\": \"{{{packageFile}}}\", \"parentDir\": \"{{{parentDir}}}\", \"depTypes\": \\[{{#each depTypes}}\"{{{this}}}\"{{#unless @last}},{{\/unless}}{{\/each}}\\]}{{#unless @last}},{{\/unless}}{{\/each}}\\]'$"
   ],
   "updateNotScheduled": false,
   "dependencyDashboard": false,


config/renovate/renovate.json
```diff
@@ -98,7 +98,13 @@
     ],
     "schedule": [
       "after 5am on saturday"
-    ]
+    ],
+    "postUpgradeTasks": {
```
gnaponie, line 102: This is the global config, and this command will be applied to all components on-boarded to Konflux. Is this intended to happen for every component? If not, putting this in a local config would be better, otherwise I'm ok with keeping this.

tkdchen: Yes, the migration tool is intended to run for all components.

gnaponie:
Gotcha. So I'm fine with this as change as it is +1

FernandesMF (approved):
Looks good to me as well!

qixiang:
/retest

re-trigger the test to check if the old buildah task can pass EC checks.

staticf0x (changes requested):

tkdchen:
Hi team, the issue has been addressed. PTAL.

staticf0x:
This doesn't unfortunately work, because the `{{{ toJSON upgrades }}}` contains apostrophes and produces
```
Command failed: pipeline-migration-tool --renovate-upgrades ' ...
/bin/sh: -c: line 1: syntax error near unexpected token `('
...
```

would it be possible to escape the apostrophes or use `encodeBase64` and modify the pipeline-migration-tool?

tkdchen:
> This doesn't unfortunately work, because the `{{{ toJSON upgrades }}}` contains apostrophes and produces
> 
> ```
> Command failed: pipeline-migration-tool --renovate-upgrades ' ...
> /bin/sh: -c: line 1: syntax error near unexpected token `('
> ...
> ```
> 
> would it be possible to escape the apostrophes or use `encodeBase64` and modify the pipeline-migration-tool?

@staticf0x How did you run the Mintmaker and get this failure? I haven't seen it happened during my test that uses a Mintmaker image with this command.

staticf0x:
Have you used the MintMaker default config? https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json it's loaded on server side and used as `RENOVATE_CONFIG_FILE`. It contains apostrophes in https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json#L92

tkdchen:
> Have you used the MintMaker default config? https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json it's loaded on server side and used as `RENOVATE_CONFIG_FILE`. It contains apostrophes in https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json#L92

Yes. I've been using my topic branch https://github.com/tkdchen/mintmaker/commits/test-STONEBLD-2824-configure-pmt/ that is based on the changes within this PR.

tkdchen:
This is the pr proposed by the migration tool
https://github.com/cqi-stonesoup-test/konflux-test-task-migrations/pull/10/files

A full Renovate log is attached, which contains the expanded command.

[renovate-job-1737430628-435df-fbkp2.log](https://github.com/user-attachments/files/18485424/renovate-job-1737430628-435df-fbkp2.log)

tkdchen:
@staticf0x PTAL

staticf0x (changes requested):
With these commas it works for me

staticf0x:
@tkdchen do you have any performance measurements of the pipeline-migration-tool? In my test, I had 14 upgrades, which resulted in 255 calls to `fetch_migration_file`, and the total run time was 7 - 9 minutes on my machine (most of it is network bound). This can be a huge issue for MintMaker, all Tekton upgrades happen on Saturday after 5 AM UTC and we have ~1000 repos to run on, unfortunately without any way to cache the results between repos.

qixiang:
> The changes for STONEBLD-3214 have been merged into the pipeline-migration-tool main branch, which has much efficient way of checking migrations for task bundles than the original. Please review this PR again. @staticf0x

Is this tool still making numerous calls to GitHub APIs? If so, are these calls authenticated or unauthenticated? I'm concerned about potential GitHub rate limiting issues.

staticf0x:
@qixiang looks like there is about ~57 requests for my 2x14 migrations (2 files) and they're to quay.io, not GitHub. @tkdchen can you confirm?

The performance is amazing now:

```
Time (mean ± σ):  	3.118 s ±  0.333 s	[User: 0.777 s, System: 0.071 s]
Range (min … max):	2.743 s …  3.378 s	3 runs
```

Down from ~96 seconds originally. Good job, @tkdchen!

I think we can merge this now.

tkdchen:
The tool relies on Quay.io only.

qixiang (approved):
> The tool relies on Quay.io only.

Thank you, then this LGTM.

staticf0x (approved):

tkdchen:
:tada:  Thank you all.

PR #136, opened 2025-01-10
Update Renovate command and set ownership

gnaponie:
Changing the placeholder command to the actual renovate command. Additionally with this commit we are setting all the necessary ownership to all the resources (ConfigMaps and secrets). So that we can be sure that one a PipelineRun is removed from the cluster all its corresponding resources are removed with it.

### DIFF
--- a/internal/pkg/tekton/pipeline_run_builder.go
+++ b/internal/pkg/tekton/pipeline_run_builder.go
@@ -106,10 +106,9 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 								TaskSpec: tektonv1.TaskSpec{
 									Steps: []tektonv1.Step{
 										{
-											Name: "renovate",
-											// TODO: use default, or get from ENV
+											Name:   "renovate",
 											Image:  "quay.io/konflux-ci/mintmaker-renovate-image:latest",
-											Script: `echo "Running Renovate"; sleep 10`,
+											Script: `RENOVATE_TOKEN=$(cat /etc/renovate/secret/renovate-token) RENOVATE_CONFIG_FILE=/etc/renovate/config/renovate.json renovate`,
 										},
 									},
 								},
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -282,14 +282,43 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(comp component.GitCo
 		return nil, err
 	}
 
-	if err := controllerutil.SetOwnerReference(pipelineRun, registry_secret, r.Scheme); err != nil {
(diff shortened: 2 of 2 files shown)


gnaponie:
@qixiang @FernandesMF @staticf0x @querti @nvtnlucie PTAL
@qixiang I've update the command to simply `renovate`, as I think it should be sufficient (the token and the config should already be present in the config map). But unfortunately I was unable to test it locally (seems like there's some issue with the private key I have configured). I'll keep digging that, but in the meantime I thought about posting the PR, if you have a stable testing environment perhaps you could already test it and review it. Thanks.

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -106,10 +106,9 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 								TaskSpec: tektonv1.TaskSpec{
 									Steps: []tektonv1.Step{
 										{
-											Name: "renovate",
-											// TODO: use default, or get from ENV
+											Name:   "renovate",
 											Image:  "quay.io/konflux-ci/mintmaker-renovate-image:latest",
-											Script: `echo "Running Renovate"; sleep 10`,
+											Script: `renovate`,
```
qixiang, line 111: You need to tell renovate where to get the token and config:
```
RENOVATE_TOKEN=$(cat /etc/renovate/secret/renovate-token) RENOVATE_CONFIG_FILE=/etc/renovate/config/renovate.json renovate
```

gnaponie:
@qixiang comment addressed. Please take another look. Thanks.

qixiang (approved):
LGTM

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -106,10 +106,9 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 								TaskSpec: tektonv1.TaskSpec{
 									Steps: []tektonv1.Step{
 										{
-											Name: "renovate",
-											// TODO: use default, or get from ENV
+											Name:   "renovate",
 											Image:  "quay.io/konflux-ci/mintmaker-renovate-image:latest",
-											Script: `echo "Running Renovate"; sleep 10`,
+											Script: `RENOVATE_TOKEN=$(cat /etc/renovate/secret/renovate-token) RENOVATE_CONFIG_FILE=/etc/renovate/config/renovate.json`,
```
qixiang, line 111: ah, renovate command is missing by accident

qixiang (changes requested):

gnaponie:
ops. I've update it, my bad. I've also tested it, and here's a nice new PR from my test run on my local machine: https://github.com/gnaponie/freshmaker/pull/13

qixiang (approved):
LGTM

PR #138, opened 2025-01-14
Support running rpm-lockfile-prototype in a subfolder

andreasgerstmayr:
Support running rpm-lockfile-prototype in a subfolder

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -310,7 +310,8 @@
   },
   "forkProcessing": "enabled",
   "allowedPostUpgradeCommands": [
-    "^rpm-lockfile-prototype rpms.in.yaml$"
+    "^rpm-lockfile-prototype rpms.in.yaml$",
+    "^rpm-lockfile-prototype [a-zA-Z0-9_-]+/rpms.in.yaml --outfile [a-zA-Z0-9_-]+/rpms.lock.yaml$"
   ],
   "updateNotScheduled": false,
   "dependencyDashboard": false


andreasgerstmayr:
Technically the dot (`.`) in the regex should be escaped, but I'm not sure if that works properly (json escaping/renovatebot regex engine) and the existing line also doesn't escape the dot.

staticf0x (changes requested):
Hello, there is currently a bug (CWFHEALTH-3864) in the `rpm` manager that prevents updating lockfiles in subdirectories (i.e. it only works for the root folder). Is there any reason this needs to be a postUpgradeCommand or would fixing that bug suffice?

andreasgerstmayr:
> Hello, there is currently a bug (CWFHEALTH-3864) in the `rpm` manager that prevents updating lockfiles in subdirectories (i.e. it only works for the root folder).

Not sure, but seems like I'm not affected by this bug.

Repository: https://github.com/os-observability/konflux-tempo
I can run `rpm-lockfile-prototype bundle-patch/rpms.in.yaml --outfile bundle-patch/rpms.lock.yaml` and it works as expected, the file `bundle-patch/rpms.lock.yaml` is created.

staticf0x:
I have implemented the bug fix I mentioned in https://github.com/redhat-exd-rebuilds/renovate/commit/8ce1bf84a209d2d1c101bd88391275013d4ea3be which should be released soon and it should fix your workflow too. The `rpm` manager will now be able to update any `rpm.lock.yaml` found in the repo. So unless there is a workflow that really requires this change, I'd close this PR, as we prefer to avoid any `postUpgradeCommands`.

andreasgerstmayr:
Ah sorry, I didn't realize the bug you referenced is about the rpm *manager*, not the `rpm-lockfile-prototype` command.
What I like about the `postUpgradeCommands` feature is that is adds the changes to the same PR which updates the base image. But probably I can achieve the same with a new `packageRule` to group these PRs?

Ok, I'm fine with closing the PR.

PR #139, opened 2025-01-14
Implement github installation token management

FernandesMF:
Implement github installation token management

This commit implements the logic to mange github installation tokens.

These tokens have a lifetime of one hour, and are limited in number. The logic to renew them responsibly is now implemented in the github module.

Other changes implemented are:
- new reconcile-timestamp label in github components
- extraction of tekton label keys in the pipelinerun controller module
- token cache in github module, to store tokens and their expiration time
- logic to cancel pipelineruns, in case there is a problem in any of launching steps

### Testing remarks
I have ran tests locally with minikube, and also in the dev cluster, and things seem to be working as expected.
- I made timing tests with many pipelineruns locally, to check if the token renewal was happening only at the times it was expected
- I made e2e tests in the dev cluster, to check if the token update in the secret is working as expected
- I made e2e tests to check if the pipelinerun cancelling logic is working fine

### DIFF
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -15,13 +15,18 @@
 package controller
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"strconv"
 
+	appstudiov1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
+	github "github.com/konflux-ci/mintmaker/internal/pkg/component/github"
 	. "github.com/konflux-ci/mintmaker/internal/pkg/constant"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/apis"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -31,8 +36,11 @@ import (
 )
 
 var (
-	MaxSimultaneousPipelineRuns = 20
-	MintMakerAppstudioLabel     = "mintmaker.appstudio.redhat.com/platform"
+	MaxSimultaneousPipelineRuns      = 20
+	MintMakerGitPlatformLabel        = "mintmaker.appstudio.redhat.com/git-platform"
+	MintMakerReconcileTimestampLabel = "mintmaker.appstudio.redhat.com/reconcile-timestamp"
+	MintMakerComponentNameLabel      = "mintmaker.appstudio.redhat.com/component"
+	MintMakerComponentNamespaceLabel = "mintmaker.appstudio.redhat.com/namespace"
 )
 
 // Collect pipelineruns with state 'running' or 'started'
@@ -59,7 +67,7 @@ type PipelineRunReconciler struct {
 
 func (r *PipelineRunReconciler) listExistingPipelineRuns(ctx context.Context, req ctrl.Request) (tektonv1.PipelineRunList, error) {
 
-	log := ctrllog.FromContext(ctx).WithName("PipelineRun")
+	log := ctrllog.FromContext(ctx).WithName("PipelineRunController")
 	var existingPipelineRuns tektonv1.PipelineRunList
 
 	err := r.Client.List(ctx, &existingPipelineRuns, client.InNamespace(req.Namespace))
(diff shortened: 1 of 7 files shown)


qixiang:
@FernandesMF we rebased `refactoring-pipelinerun` branch and force-pushed it, could you rebase this PR?

After generated the new token, you have to update the secret to store the new token in it, otherwise the PipelineRun won't be able to consume this new token.

FernandesMF:
Hey team, don't worry about the rebase! I pulled the new changes into this branch.

ALso good catch on the secret! I implemented the update of the new token, please have a look and double check if I'm doing it correctly. I used base64 encoding, but I was not completely sure if it is needed.

internal/pkg/component/github/github.go
```diff
@@ -23,14 +25,14 @@ import (
...
-	ghAppInstallationTokenCacheMutex sync.Mutex
-	ghAppInstallationTokenCacheID    int64
-	ghAppID                          int64
-	ghAppPrivateKey                  []byte
+	ghAppInstallationsCache              *Cache
+	ghAppInstallationsCacheMutex         sync.Mutex
+	ghAppInstallationsCacheID            int64
+	ghAppInstallationTokenCache          TokenCache
+	ghAppInstallationTokenTimeThreshold  = 30 * time.Minute
+	ghAppInstallationTokenValidityWindow = time.Hour
```
qixiang, line 33: Just use `time.hour` is easier than the long name, it can never been changed, so unnecessary to define a variable.

internal/pkg/component/github/github.go
```diff
@@ -46,6 +48,16 @@ type Component struct {
...
+type TokenInfo struct {
+	Token     string
+	ExpiresAt time.Time
+}
+
+type TokenCache struct {
+	mu      sync.RWMutex
+	entries map[string]TokenInfo
+}
+
```
qixiang, line 60: You can move this to cache.go file.

internal/pkg/component/github/github.go
```diff
@@ -106,6 +118,29 @@ func NewComponent(comp *appstudiov1alpha1.Component, timestamp int64, client cli
...
+	}
+
+	// when token is going to expire in 30m, we can't use it
+	if time.Until(entry.ExpiresAt) < ghAppInstallationTokenTimeThreshold {
+		return TokenInfo{}, false
+	}
+
+	return entry, true
+}
+
```
qixiang, line 143: As above, you can move it to cache.go.

internal/pkg/component/github/github.go
```diff
@@ -118,20 +153,7 @@ func (c *Component) GetBranch() (string, error) {
...
-		ghAppInstallationTokenCache = NewCache()
-		ghAppInstallationTokenCacheID = c.Timestamp
-	}
-
-	appInstallations, err := c.getAppInstallations()
-	if err != nil {
-		return "", fmt.Errorf("failed to get GitHub App installations: %w", err)
-	}
-
+func (c *Component) getMyInstallationID(appInstallations []AppInstallation) (int64, error) {
```
qixiang, line 156:
```suggestion
func (c *Component) getInstallationID() (int64, error) {
```
No necessary to pass the installations, you can call `c.getAppInstallations()` inside the function directly.

internal/pkg/component/github/github.go
```diff
@@ -149,13 +171,47 @@ func (c *Component) GetToken() (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	ghAppInstallationTokenCache.Set(tokenKey, tokenInfo)
 
-	token_key := fmt.Sprintf("installation_%d", installationID)
-	if token, ok := ghAppInstallationTokenCache.Get(token_key); ok {
-		return token.(string), nil
+	appSecret := corev1.Secret{}
+	appSecretKey := types.NamespacedName{Namespace: "mintmaker", Name: "pipelines-as-code-secret"}
```
qixiang, line 204: This secret is the secret storing GitHub App ID and private key, not the secret used by the github component.

internal/pkg/component/github/github.go
```diff
@@ -149,13 +171,47 @@ func (c *Component) GetToken() (string, error) {
...
+	err = c.client.Get(c.ctx, appSecretKey, &appSecret)
+	if err != nil {
+		return "", err
 	}
+	appSecret.Data["github-private-key"] = []byte(base64.StdEncoding.EncodeToString([]byte(tokenInfo.Token)))
+
+	return tokenInfo.Token, nil
+}
+
+func (c *Component) generateNewToken(installationID int64) (TokenInfo, error) {
```
qixiang, line 214: There is no necessary to add a new API, you can refactor `GetToken` to support checking expiring time when necessary.

internal/pkg/component/github/github.go
FernandesMF: I searched the codebase and had a look at the deployed secrets, but I confess I couldn't figure by myself what the correct secret is 😅 
Could you point me to the correct one, please?

qixiang: This is the secret:
https://github.com/konflux-ci/mintmaker/blob/062080578747badc43ab7b74db501f9e429901f8/internal/controller/dependencyupdatecheck_controller.go#L201-L218
https://github.com/konflux-ci/mintmaker/blob/062080578747badc43ab7b74db501f9e429901f8/internal/controller/dependencyupdatecheck_controller.go#L238-L245

internal/pkg/component/github/cache.go
```diff
@@ -19,3 +22,36 @@ func (c *Cache) Get(key string) (interface{}, bool) {
...
+
+type TokenCache struct {
+	mu      sync.RWMutex
+	entries map[string]TokenInfo
+}
+
+func (c *TokenCache) Set(key string, tokenInfo TokenInfo) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.entries[key] = tokenInfo
```
gnaponie, line 39: Hey Mario. This is the line that causes an error in my test env.
The issue is that this `c.entries` is nil when you assign this `tokenInfo`.  So it crashes the whole controller.
I'm unable to see exactly how to fix that now. But if I comment this line (which is probably not the best solution), the error doesn't happen. So you might want to change this line. I hope it helps.

FernandesMF: Awesome, thanks for spotting that, Giulia!

My accustomed-to-python-mind didn't realize at first that `TokenCache.entries` was only declared, but never initialized 😅 . Then the first time this assignment is ran (L39), it will fail because `c.entries` is still a the nil map...

I made a fix to the `Set` function that will allocate the map the first time it is run. I also wrote a silly unit test to make sure the `assignment to entry in nil map` error is duely fixed.

internal/controller/pipelinerun_controller.go
```diff
@@ -95,15 +102,52 @@ func (r *PipelineRunReconciler) launchUpToNPipelineRuns(numToLaunch int, existin
...
+				appSecret := corev1.Secret{}
+				appSecretKey := types.NamespacedName{Namespace: "mintmaker", Name: pipelineRun.Name}
+				err = r.Client.Get(ctx, appSecretKey, &appSecret)
+				if err != nil {
+					return err
+				}
+				tokenInfo := github.TokenInfo{
+					Token: token, ExpiresAt: time.Now().Add(GhTokenValidity),
+				}
+				appSecret.Data["renovate-token"] = []byte(base64.StdEncoding.EncodeToString([]byte(tokenInfo.Token)))
```
qixiang, line 137: I don't see where the secret is being updated. You need to call update/patch to make the change.

internal/pkg/component/github/cache.go
```diff
@@ -19,3 +22,42 @@ func (c *Cache) Get(key string) (interface{}, bool) {
...
+}
+
+func (c *TokenCache) Set(key string, tokenInfo TokenInfo) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// initialize entries map in the first usage
+	if c.entries == nil {
+		c.entries = make(map[string]TokenInfo)
+	}
```
qixiang, line 43: Instead of initialization the member of the struct, you should make sure the TokenCache instance is created, in a away of:

```
var myTokenCache *TokenCache

func someFunc(...) {

   if myTokenCache == nil {
       myTokenCache = &TokenCache{
         ...
         entries: make(map[string]TokenInfo)
       }
   }
```

FernandesMF: Good idea! I made an `init` function in the module and moved the initialization there

internal/controller/pipelinerun_controller.go
```diff
@@ -95,15 +102,52 @@ func (r *PipelineRunReconciler) launchUpToNPipelineRuns(numToLaunch int, existin
...
+				}
+				appSecret := corev1.Secret{}
+				appSecretKey := types.NamespacedName{Namespace: "mintmaker", Name: pipelineRun.Name}
+				err = r.Client.Get(ctx, appSecretKey, &appSecret)
+				if err != nil {
+					return err
+				}
+				tokenInfo := github.TokenInfo{
+					Token: token, ExpiresAt: time.Now().Add(GhTokenValidity),
+				}
```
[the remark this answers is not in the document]
FernandesMF: I thought that we also needed to update the token validity time. That is why I used a `TokenInfo` object and updated both fields. Is that not correct?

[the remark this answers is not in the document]
gnaponie: I think what Qixiang means is that you've declared a variable `tokenInfo`, but then you've used it just once in the following line. And you could have just put

``github.TokenInfo{Token: token, ExpiresAt: time.Now().Add(GhTokenValidity),}``

instead of `tokenInfo`, and you'd avoid declaring another variable. But I personally think it's ok if you want to keep `tokenInfo`, the code is easier to read this way. This isn't a blocking comment.

gnaponie:
This looks correct to me and I was able to test it locally successfully. +1
But due to the impact of this change I'd also wait for Qixiang's approval if you don't mind.

gnaponie (approved):

FernandesMF:
Hi team! During my local tests, I found that the time checks of the token validity were not quite right. I also changed the code to update the token in the secret.

Could you please have another look?

Thanks!

internal/controller/pipelinerun_controller.go
```diff
@@ -95,21 +104,70 @@ func (r *PipelineRunReconciler) launchUpToNPipelineRuns(numToLaunch int, existin
...
 			original := pipelineRun.DeepCopy()
+
+			// renew github secret if it is no more valid or too old
+			if pipelineRun.Labels[MintMakerGitPlatformLabel] == "github" {
+				var component appstudiov1alpha1.Component
+				var ghComponent *github.Component
+				componentName := pipelineRun.Labels[MintMakerComponentLabel]
+				componentKey := types.NamespacedName{Namespace: "mintmaker", Name: componentName}
+				err = r.Client.Get(ctx, componentKey, &component)
+				if err != nil {
```
qixiang, line 115: If any error happens, we should continue to next PipelineRun instead of returning.

internal/pkg/component/github/github.go
```diff
@@ -46,6 +47,12 @@ type Component struct {
 	ctx           context.Context
 }
 
+func init() {
+	ghAppInstallationTokenCache = TokenCache{
+		mu: sync.RWMutex{}, entries: make(map[string]TokenInfo),
+	}
+}
+
```
qixiang, line 55: It's unnecessary to initialize the cache in `init`, you can do that in `GetToken`.

internal/controller/pipelinerun_controller.go
FernandesMF: Good catch!
I changed the `return`s to `continue`s inside the loop.
I also added lines to log the cause of the skip in each case.

internal/pkg/component/github/github.go
FernandesMF: Thanks for the input!

I think that leaving the initialization has possibly two advantages: 
- it might be the most intuitive option (since it is an initialization task, going in the initialization function)
- it refrains from adding another responsibility to the `GetToken` function, and is more in line with the 'single responsibility principle'

Could you explain further, why you think it is unnecessary to use the `init` function, and why the initialization belongs better in `GetToken`?

internal/controller/pipelinerun_controller.go
```diff
@@ -95,21 +104,74 @@ func (r *PipelineRunReconciler) launchUpToNPipelineRuns(numToLaunch int, existin
...
+					log.Error(err, "Unable to update fetch konflux component, skipping")
+					continue
```
qixiang, line 117: I realized just skipping and continue to next pipelinerun is not sufficient, we'll leave these pipelineruns in pending status forever (they won't be garbage collected), and this controller will process them again, resulting in infinite loop.

internal/pkg/component/github/github.go
qixiang: Basically `init` should be avoided unless you have some special reasons, you can search some articles about this. In our case, it's straightforward to initialize the cache in GetToken, and you have already done that, so we don't need to do that in `init`.

qixiang (changes requested):

internal/pkg/component/github/github.go
FernandesMF: I have searched a bit and now I understand the bad aspects of using `init`, they didn't occur to my at first. Thank you for the explanation.

I have removed `init`, and moved the initialization to `GetToken`.

internal/controller/pipelinerun_controller.go
FernandesMF: I made slight changes to the code and now it is cancelling the pipelineruns instead of just skipping them, in case there is a problem in any of the launching steps.

internal/controller/pipelinerun_controller.go
```diff
@@ -74,45 +81,144 @@ func filterPipelineRunListWithLabel(pipelineRunList tektonv1.PipelineRunList, la
...
+	}
+
+	// proceed with fetching intermediate steps to check token, in case of a github pipelinerun
+	var err error
+	var message string
+	var component appstudiov1alpha1.Component
+	var ghComponent *github.Component
+
+	componentName := plr.Labels[MintMakerComponentLabel]
+	componentKey := types.NamespacedName{Namespace: "mintmaker", Name: componentName}
```
qixiang, line 132: Components are not in `mintmaker `namespace.

FernandesMF: This was really going unnoticed, thanks for your attention!
I adjusted the dependencyupdatecheck controller to store the component's namespace in the labels, when creating the pipelinerun, and the pipelinerun controller to fetch it there.

qixiang (approved):
LGTM

FernandesMF:
Deployment and test in stage were successful, I'll merge this PR.

PR #140, opened 2025-01-15
Enable pip-compile as allowedPostUpgradeCommand

Zokormazo:
We have a use case where we want to trigger a pip-compile requirements.txt file refresh as a post upgrafe command. The regex is flexible to match and arbitraty number of flags while calling pip-compile

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -310,7 +310,8 @@
   },
   "forkProcessing": "enabled",
   "allowedPostUpgradeCommands": [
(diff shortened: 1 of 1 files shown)


staticf0x:
Can you please point us to the workflow that requires this change? We really don't like to allow anything in `allowedPostUpgradeCommands` unless it's absolutely necessary.

staticf0x:
Please use CWFHEALTH-3924 to provide a description of your workflow or check if something similar isn't proposed already. I will have to close this PR now for the reasons mentioned in my above comment and in the mentioned ticket.

PR #141, opened 2025-01-15
Determine component's git repo default branch

gnaponie:
This used to hard-coded to main, let's query github and gitlab API to get the correct default branch.


gnaponie:
@qixiang @FernandesMF PTAL
This runs correctly, but I didn't really test it yet. But it's ready for a review.

internal/pkg/component/gitlab/gitlab.go
```diff
@@ -4,12 +4,15 @@ import (
...
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 
 	appstudiov1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
 	"github.com/konflux-ci/mintmaker/internal/pkg/component/base"
 	bslices "github.com/konflux-ci/mintmaker/internal/pkg/slices"
 	"github.com/konflux-ci/mintmaker/internal/pkg/utils"
+	gitlab_mintmaker "github.com/konflux-ci/mintmaker/pkg/git/gitlab"
```
qixiang, line 14: we're going to remove the `pkg` directory (not the `internal/pkg`) as a code clean up step in this refactoring work, could you move the necessary APIs to `internal/pkg/component/gitlab`?

gnaponie: good point. Rebased.

internal/pkg/component/gitlab/gitlab.go
```diff
@@ -168,9 +194,55 @@ func (c *Component) GetAPIEndpoint() string {
...
+}
+
 func (c *Component) getDefaultBranch() (string, error) {
-	// TODO: call github APIs to determine the default branch
-	return "main", nil
+	token, err := c.GetToken()
+	if err != nil {
+		return "", fmt.Errorf("failed to get GitLab token: %w", err)
+	}
+	baseUrl, err := c.GetBaseUrl(c.Repository)
```
qixiang, line 232: This might be misleading, but `c.Repository` is actually the repo path, not the git URL: https://github.com/konflux-ci/mintmaker/blob/fc51f926095d3ba47b343d56a9d7951b4b035bee/internal/pkg/component/base/base.go#L30-L32

internal/pkg/component/gitlab/gitlab.go
```diff
@@ -28,6 +30,31 @@ type Repository struct {
...
+}
+
+type MissingHostError struct {
+	url string
+}
+
+func (e MissingHostError) Error() string {
+	return fmt.Sprintf("Failed to detect host in url %s", e.url)
+}
+
```
qixiang, line 57: We don't need such checks, the url has been validated already while creating the component, such as: https://github.com/konflux-ci/mintmaker/blob/fc51f926095d3ba47b343d56a9d7951b4b035bee/internal/pkg/component/gitlab/gitlab.go#L33-L45

qixiang (changes requested):
Regarding the gitlab implementation, we have some data stored in the component already, we can use that directly, e.g.:
```
func (c *Component) getDefaultBranch() (string, error) {
	token, err := c.GetToken()
	if err != nil {
		return "", fmt.Errorf("failed to get GitLab token: %w", err)
	}

	u, err := url.Parse(c.GitURL)
	if err != nil {
		return "", fmt.Errorf("failed to parse git url: %w", err)
	}
	baseURL := u.Scheme + "://" + c.Host
	client, _ := gitlab.NewClient(token, gitlab.WithBaseURL(baseURL))
	project, _, err := client.Projects.GetProject(c.Repository, nil)
	if err != nil {
		return "", err
	}
	if project == nil {
		return "", fmt.Errorf("project info is empty in GitLab API response")
	}
	return project.DefaultBranch, nil
}
```
Could you rebase this PR on top of latest `refactoring-pipelinerun` branch? which will be more convenient for testing locally (it's missing a fix in https://github.com/konflux-ci/mintmaker/pull/136)?

qixiang:
Another thing is the go.mod should be updated, because new packages are introduced.

gnaponie:
Comments addressed.

qixiang (approved):
LGTM

PR #151, opened 2025-01-20
Allow controller to update pipelineruns

qixiang:
Update role permissions to enable controller service account to update pipelinerun resources


qixiang:
@gnaponie @FernandesMF we forgot to update role to allow our controller to update PipelineRuns.

gnaponie (approved):

PR #159, opened 2025-01-22
Shorten name due to char limits on some objects

gnaponie:
It seems like for some objects we are getting some char limit restrictions. Even though having the component in the name of the PipelineRun and in the config map might be useful for debugging reasons, we cannot really be sure that it won't contain a too long character. Due to this reason, let's remove the component name.

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -175,7 +175,7 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(comp component.GitCo
 
 	log := ctrllog.FromContext(ctx).WithName("DependencyUpdateCheckController")
 	ctx = ctrllog.IntoContext(ctx, log)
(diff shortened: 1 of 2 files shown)


gnaponie:
@qixiang it's a very nice idea to have the component name, but I guess we can never be sure that the component doesn't have a too long name. The error seems to be mostly with the config map in the task, perhaps we can avoid the component name only there, instead that in all the pipelineruns?

qixiang (approved):
LGTM
Seems many k8s resources and fields have a maximum length limit of 63 characters, before we figure out a better solution for convenience, let's just use label to filter PipelineRuns.

qixiang:
btw, if you have a plan to deploy this to stage, could you also change the maximum number of pipelinerun? so we can check if we're unable to run maybe 20 pipelineruns at the same time in stage.

gnaponie:
Yes, let me do that in another commit of this PR perhaps. I'll do that in a bit. Then I'll try to test this locally first, and then we can promote this to stage.

qixiang:
LGTM

PR #160, opened 2025-01-23
Remove ownership for the caConfigMap

gnaponie:
The ca-bundle configmap should not be owned by any pipelinerun, as it is shared by all PipelineRuns.


qixiang (approved):
LGTM

staticf0x (approved):

PR #161, opened 2025-01-23
Cleanup resources if any error occurs during PipelineRun creation

qixiang:
Previously, if any error occurred after ConfigMaps/Secrets were created but before setting their ownership to the PipelineRun, these resources would remain orphaned in the cluster. For example, if the creation of PipelineRun failed, all the configmaps and secrets created for this PipelineRun won't be deleted.

This change ensures proper cleanup by deleting any created resources when errors occur at any point in the workflow.

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -175,10 +175,25 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(comp component.GitCo
 
 	log := ctrllog.FromContext(ctx).WithName("DependencyUpdateCheckController")
 	ctx = ctrllog.IntoContext(ctx, log)
-	name := fmt.Sprintf("renovate-%d-%s", comp.GetTimestamp(), RandomString(8))
-	registry_secret, err := r.createMergedPullSecret(ctx)
 
-	renovateConfig, err := comp.GetRenovateConfig(registry_secret)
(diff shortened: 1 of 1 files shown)


internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -310,6 +328,7 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(comp component.GitCo
 		return nil, err
 	}
 
+	resources = nil
```
staticf0x, line 331: Is this intended? I thought `defer` ran just before `return`, so will there be anything to clean up?

gnaponie (approved):

gnaponie:
Looks good to me.

internal/controller/dependencyupdatecheck_controller.go
qixiang: If the code can reach here, there was no error happened, means we should not delete any resources we have created in this function, so we assign `nil` to resource, then the defer func will have nothing to delete.

qixiang: whenever we created a resource, add it to `resources`, if any error happens, all these resources will not be owned by the pipelinerun, they should should be deleted. If no happen happens, don't delete them, because they will be used and owned by the pipelinerun, they will be deleted automatically when the pipelinerun is deleted, or garbage collected.

PR #89, opened 2024-10-30
Convert Red Hat CSAF VEX files to OSV database format

nvtnlucie:
Red Hat publishes vulnerability data in CSAF VEX format. To facilitate easier consumption of this data by Renovate, this PR converts these CSAF VEX files into OSV database format. The script takes an url as input (from https://security.access.redhat.com/data/csaf/v2/advisories/2024/), and outputs a .nedb file containing osv vulnerability data (with the same fields as https://github.com/renovatebot/osv-offline/releases , except in the scope of this task the file contains vulnerabilities from one advisory only). There is a `main()` function used to demonstrate how it works, which will be deleted once [CWFHEALTH-3440](https://issues.redhat.com/browse/CWFHEALTH-3440) is implemented. These outputs were tested to validate against https://raw.githubusercontent.com/ossf/osv-schema/main/validation/schema.json .

The script also ignores everything which is not a rpm dependency. I am not sure about the name (or placement) of this module yet, I am accepting suggestions :).

Closes [CWFHEALTH-3757](https://issues.redhat.com/browse/CWFHEALTH-3757).

PR #163, opened 2025-01-24
Introduce a script to generate a database usable by osv-offline

querti:
This script largely builds on top of PR #89 with some necessary modifications and additions.

The finalized script downloads and extracts the upstream OSV database and generates additional files for container and RPM vulnerabilities.

Few modifications to the original scripts were necessary:
- Add a retry mechanism for getting the CSAF VEX files. Transient networking errors were common during testing and retries shold resolve them.
- Add internal ID (_id) to the OSV schema and set it to a random string. Every object in the nedb database must have this property.
- Split how the affectedList is generated between RPMs and containers.
- Duplicate affectedList entries for containers to also have registry registry.access.redhat.com. This is necessary because the CSAF VEX data only tracks images with registry.redhat.io.

This script is expected to be run periodically by a cronjob and regenerate the DB files onto a persistent volume. Add the new script to the mintmaker container, so that the cronjob can use the same image.

WARNING: The script now gets the list of advisories from changes.csv. This is incorrect and should be changed to releases.csv when it becomes available. The current state is for testing purposes only.

### DIFF
--- a/cmd/osv-generator/main.go
+++ b/cmd/osv-generator/main.go
@@ -25,10 +25,10 @@ import (
 )
 
 func main() {
-	dockerFilename := flag.String("docker-filename", "docker.nedb", "Filename for the Docker DB file")
+	containerFilename := flag.String("container-filename", "docker.nedb", "Filename for the Container DB file")
 	rpmFilename := flag.String("rpm-filename", "rpm.nedb", "Filename for the RPM DB file")
 	destDir := flag.String("destination-dir", "/tmp/osv-offline", "Destination directory for the OSV DB files")
-	days := flag.Int("days", 90, "Only advisories created in the last X days are included")
+	days := flag.Int("days", 120, "Only advisories created in the last X days are included")
 
 	flag.Parse()
 	err := os.MkdirAll(*destDir, 0755)
@@ -43,7 +43,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	osv_generator.GenerateOSV(filepath.Join(*destDir, *dockerFilename), true, *days)
+	osv_generator.GenerateOSV(filepath.Join(*destDir, *containerFilename), true, *days)
 	if err != nil {
 		fmt.Println("Generating the container OSV database has failed: ", err)
 		os.Exit(1)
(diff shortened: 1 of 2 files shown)


staticf0x:
Thank you this looks pretty neat!

nvtnlucie (dismissed):
LGTM except for one thing, good job on adding the retry mechanism :+1:  !

nvtnlucie (dismissed):

cmd/osv-generator/main.go
```diff
@@ -16,14 +16,42 @@ package main
...
+	"path/filepath"
 
+	osv_downloader "github.com/konflux-ci/mintmaker/tools/osv-downloader"
 	osv_generator "github.com/konflux-ci/mintmaker/tools/osv-generator"
 )
 
-// A demo which parses RPM CVE data created in the last 24 hours into OSV database format
 func main() {
-	filename := flag.String("filename", "redhat.nedb", "Output filename for OSV database")
+	dockerFilename := flag.String("docker-filename", "docker.nedb", "Filename for the Docker DB file")
```
qixiang, line 28: we should try use `container` instead of `docker` as possible as we can.

tools/osv-generator/generator.go
```diff
@@ -26,9 +26,8 @@ import (
 const URL = "https://security.access.redhat.com/data/csaf/v2/advisories"
 
 // Get list of new advisories published in the last `days`
-// TODO: Do not use until published.csv is periodically generated
 func getAdvisoryListByPublished(days int) ([]string, error) {
```
qixiang, line 29: Since we're using `releases.csv` instead of `published.csv`, how about renaming this to `getAdvisoryListByReleases`?

tools/osv-generator/generator.go
```diff
@@ -26,9 +26,8 @@ import (
 const URL = "https://security.access.redhat.com/data/csaf/v2/advisories"
 
 // Get list of new advisories published in the last `days`
```
qixiang, line 28:
```suggestion
// Get a list of advisory filenames from releases.csv that were released
// within the specified number of days from the current time. The advisory
// filenames in releases.csv are sorted by released date, newest first.
```

qixiang (approved):

PR #177, opened 2025-01-30
Set resources for the PipelineRun

gnaponie:
This way we:
- Make sure the pod gets the resources it needs and it won't get killed.
- We help OCP scheduler spread the load on the worker nodes.

### DIFF
--- a/internal/pkg/tekton/pipeline_run_builder.go
+++ b/internal/pkg/tekton/pipeline_run_builder.go
@@ -23,6 +23,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


staticf0x (approved):

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -118,6 +119,16 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 													Type: corev1.SeccompProfileTypeRuntimeDefault,
 												},
 											},
+											ComputeResources: corev1.ResourceRequirements{
+												Limits: corev1.ResourceList{
+													"cpu":    resource.MustParse("150m"),
+													"memory": resource.MustParse("512Mi"),
```
hugares, line 125: The limits and requests are mixed up, requests should be lower or equal to the limits

staticf0x (dismissed):

internal/pkg/tekton/pipeline_run_builder.go
gnaponie: Ops! I inverted them, apologies. I've addressed the comment. May you have another look? Thanks.

hugares (commented):
LGTM

staticf0x (approved):

PR #179, opened 2025-02-03
Set GOTOOLCHAIN for Renovate's child processes

staticf0x:
Docs: https://docs.renovatebot.com/self-hosted-configuration/#customenvvariables

I was confused why some env variables were available to Renovate and some not. Turns out, some are hardcoded ([env.ts](https://github.com/renovatebot/renovate/blob/main/lib/util/exec/env.ts)) to be inherited from outside and the rest needs to go in `customEnvVariables`. This seems like a good protection against leaking secrets.

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -13,6 +13,9 @@
   "inheritConfig": true,
   "platformCommit": "enabled",
   "autodiscover": false,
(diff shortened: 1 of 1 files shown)


nvtnlucie (approved):
LGTM!

PR #181, opened 2025-02-05
Get renovate bot id dynamically

nvtnlucie:
The bot id was previously hardcoded into the string which caused a bug at one of the clusters. The cluster uses a different bot users and renovate then cant recognize its own PRs and stops updating them.

Closes [CWFHEALTH-3955](https://issues.redhat.com/browse/CWFHEALTH-3955) -> [CWFHEALTH-3954](https://issues.redhat.com/browse/CWFHEALTH-3954)

### DIFF
--- a/internal/pkg/component/github/github.go
+++ b/internal/pkg/component/github/github.go
@@ -305,6 +305,18 @@ func (c *Component) getAppSlug() (string, error) {
 	return slug, nil
 }
 
+func (c *Component) getUserId(username string) (int64, error) {
+	// No need to add auth here as User API is public
+	client := github.NewClient(&http.Client{})
+
+	user, _, err := client.Users.Get(context.Background(), username)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get user information: %w", err)
+	}
+	id := user.GetID()
+	return id, nil
+}
+
 func (c *Component) GetRenovateConfig(registrySecret *corev1.Secret) (string, error) {
 	baseConfig, err := c.GetRenovateBaseConfig(c.client, c.ctx, registrySecret)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


staticf0x:
Thank you for this, it looks very good! I have one question: how does it behave on repeated calls? Is the result cached in the client?

nvtnlucie:
> Thank you for this, it looks very good! I have one question: how does it behave on repeated calls? Is the result cached in the client?

I will take a look, during implementation i noticed we create the client like 5 times. Some calls cant be shared between components, but maybe there is room for optimization

nvtnlucie:
Ok, it is cached even if you create a new client and request the same user.

staticf0x (dismissed):

qixiang:
> Ok, it is cached even if you create a new client and request the same user.

Seems this is not true, if you check the resp returned from github, it includes the rate limit info, every time this API is called, the remaining value is decreased by 1.
```
user, resp, err := client.Users.Get(context.Background(), username)
if resp != nil {
    fmt.Printf("Rate limit remaining: %d\n", resp.Rate.Remaining)
}
```

nvtnlucie:
> > Ok, it is cached even if you create a new client and request the same user.
> 
> Seems this is not true, if you check the resp returned from github, it includes the rate limit info, every time this API is called, the remaining value is decreased by 1.
> 
> ```
> user, resp, err := client.Users.Get(context.Background(), username)
> if resp != nil {
>     fmt.Printf("Rate limit remaining: %d\n", resp.Rate.Remaining)
> }
> ```

Thank you for noticing this, i tested it only by checking that subsequent call are 10x faster, had no idea about Rate.

internal/pkg/component/github/github.go
```diff
@@ -305,6 +307,31 @@ func (c *Component) getAppSlug() (string, error) {
 	return slug, nil
 }
 
+func (c *Component) getUserId(username string) (int64, error) {
```
qixiang, line 310: This user doesn't change throughout the controller's lifecycle (if the github app is changed, we need to change the github app data in secret and restart controller), I think using a cache is unnecessary, a package-level variable should be sufficient. Similar to `ghAppID` and `ghAppPrivateKey`.
```
var ghAppUserID int64
...
if ghAppUserID != 0 {
   return ghAppUserID, nil
# call API to get the user ID.
```

nvtnlucie: Okay, third time's the charm :)

qixiang (approved):
LGTM, please squash these commits while merging it.

staticf0x (approved):

PR #189, opened 2025-02-12
Build and add osv-generator binary to image

qixiang:
- Build osv-generator binary alongside manager binary
- Copy osv-generator binary to the final controller image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,19 +12,23 @@ COPY go.sum go.sum
 RUN go mod download
 
 # Copy the go source
-COPY cmd/manager/main.go cmd/main.go
+COPY cmd/manager/main.go cmd/manager/main.go
+COPY cmd/osv-generator/main.go cmd/osv-generator/main.go
 COPY api/ api/
+COPY tools/ tools/
 COPY internal/ internal/
 COPY licenses/ licenses/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager cmd/main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager cmd/manager/main.go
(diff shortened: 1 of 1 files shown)


staticf0x (approved):

qixiang:
EC check failed with error:
```
[Violation] java.trusted_dependencies_source_list_provided
  ImageRef: quay.io/redhat-user-workloads/konflux-mintmaker-tenant/mintmaker/mintmaker@sha256:10233191471e9d31548c7d1d3df92b469296148357fb1139f57a1cb0ad77de44
  Reason: Rule data allowed_java_component_sources has unexpected format: (Root): Array must have at least 1 items
  Title: Trusted Java dependency source list was provided
  Description: Confirm the `allowed_java_component_sources` rule data was provided, since it's required by the policy rules in
  this package. To exclude this rule add "java.trusted_dependencies_source_list_provided" to the `exclude` section of the policy
  configuration.
  Solution: Add a data source that contains allowable source repositories for build dependencies. The source must be located under
  a key named 'allowed_java_component_sources'. More information on adding xref:ec-cli:ROOT:configuration.adoc#_data_sources[data
  sources].
```
This is caused by a recent change in EC policy, I'll merge this PR after the issue is fixed in https://github.com/release-engineering/rhtap-ec-policy/pull/112.

qixiang:
> May I ask what's the point of the PR? I thought we decided to create a new image for the DB.

Do you want to build the binary in that database image? I think it's much easier to build the binary here because we already has the sources here. After this change, you can just copy the binary and run it to generate the database. For example:
```
FROM quay.io/konflux-ci/mintmaker:latest as tools

FROM registry.access.redhat.com/ubi9/ubi-minimal:9.5@sha256:b87097994ed62fbf1de70bc75debe8dacf3ea6e00dd577d74503ef66452c59d6
WORKDIR /tmp
COPY --from=tools /osv-generator /
RUN /osv-generator
```

qixiang:
This will be addressed by https://github.com/konflux-ci/mintmaker/pull/163, closing.

PR #191, opened 2025-02-12
Introduce OSV database to the PipelineRuns

querti:
PipelineRuns will now have two tasks and a shared persistent volume. The first task copies the OSV DB from the periodically built DB image to the shared volume. The second task runs renovate and can utilize the OSV database in its run. New env variables were added to specify the DB location and disable the download of the upstream DB. The created persistent volume is deleted along with the PipelineRun.

I have increased the memory requirements to 2Gi because I ran into OOM issues when testing with 1Gi of memory. The reason is most likely that the DB is loaded into memory during execution, which increases the memory foorprint by up to ~500MB. It's possible that 2Gi is an overkill, but I haven't tested the safe threshold more closely.

### DIFF
--- a/internal/pkg/tekton/pipeline_run_builder.go
+++ b/internal/pkg/tekton/pipeline_run_builder.go
@@ -107,12 +107,41 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 			Spec: tektonv1.PipelineRunSpec{
 				Status: tektonv1.PipelineRunSpecStatusPending,
 				PipelineSpec: &tektonv1.PipelineSpec{
+					Workspaces: []tektonv1.PipelineWorkspaceDeclaration{
+						{
+							Name: "shared-db",
+						},
+					},
 					Tasks: []tektonv1.PipelineTask{
 						{
 							Name: "build",
+							Workspaces: []tektonv1.WorkspacePipelineTaskBinding{
+								{
+									Name:      "shared-db",
+									Workspace: "shared-db",
+								},
+							},
 							TaskSpec: &tektonv1.EmbeddedTask{
 								TaskSpec: tektonv1.TaskSpec{
+									Workspaces: []tektonv1.WorkspaceDeclaration{
+										{
+											Name: "shared-db",
+										},
+									},
 									Steps: []tektonv1.Step{
+										{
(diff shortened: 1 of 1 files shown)


staticf0x (dismissed):

qixiang:
@querti can we add a new step in task instead of a new task? iirc, all steps in one task will be executed in one pod, and they're executed sequentially in the order they are defined (you don't need to use `runAfter`).

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -156,6 +216,23 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 						},
 					},
 				},
+				Workspaces: []tektonv1.WorkspaceBinding{
+					{
+						Name: "shared-db",
+						VolumeClaimTemplate: &corev1.PersistentVolumeClaim{
```
qixiang, line 222: The data only need to be shared within this PipelineRun, an emptyDir volume is sufficient.

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -127,12 +156,12 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 											},
 											ComputeResources: corev1.ResourceRequirements{
 												Requests: corev1.ResourceList{
-													"cpu":    resource.MustParse("150m"),
-													"memory": resource.MustParse("512Mi"),
+													"cpu":    resource.MustParse("300m"),
+													"memory": resource.MustParse("1Gi"),
```
qixiang, line 160: Is increasing `requests.limits` sufficient for us? And since the limit of cpu is still `300m`, I don't think it makes sense to increase `cpu` request to `300m`.

qixiang (commented):
LGTM, just a minor comment. We're not ready for merging this yet, right? At least the database image is not released to destination repo.

internal/pkg/tekton/pipeline_run_builder.go
qixiang: How about `requests.memory`? Is original `512Mi` sufficient as the minimal requirement for pod to start?

qixiang (approved):
LGTM

PR #193, opened 2025-02-13
Update the release process doc

staticf0x:
Since we now have semi-automated releases, I thought we could mention that in the docs.


qixiang (approved):
LGTM

I filed CWFHEALTH-3998 to further improve the doc about the release process through infra-deployment.

PR #197, opened 2025-02-17
Enable gitlabci manager

staticf0x:
We can safely enable these two managers since we just got a user last week with successful runs and I also added E2E tests for these.


qixiang (approved):
:+1:

PR #205, opened 2025-02-25
Cwfhealth 3293

FernandesMF:
This commit adds tests to the pipelinerun- and dependencyupdatecheck-controllers, which were implemented as part of the modularization effort.

### Testing remarks
The tests are passing with `make tests`. Some of them have intermittent failures, but I allowed those to be flaked in the test suites, so it should be fine.


internal/controller/pipelinerun_controller_test.go
```diff
@@ -72,58 +88,111 @@ var _ = Describe("PipelineRun Controller", FlakeAttempts(3), func() {
...
+				types.NamespacedName{
+					Name: labels[MintMakerComponentNameLabel], Namespace: labels[MintMakerComponentNamespaceLabel],
+				}, "app", "https://github.com/testcomp.git", "gitrevision", "gitsourcecontext",
+			)
+
+			secretData := map[string]string{
+				"github-application-id": "1234567890",
+				"github-private-key":    testPrivateKey,
+				"webhook.secret":        "webhook.secret",
+				"webhook.url":           "webhook.url",
```
qixiang, line 120: Can remove these two key-values.

PR #217, opened 2025-03-06
Remove kube-rbac-proxy

qixiang:
Remove kube-rbac-proxy per
https://github.com/kubernetes-sigs/kubebuilder/discussions/3907


qixiang:
This closes #110.

This change was primarily implemented by upgrading to a new version of Kubebuilder and regenerating scaffold code, then comparing differences between the new and old code to remove outdated portions. As a result, this change introduces numerous unrelated modifications that came with the new Kubebuilder version.

staticf0x (approved):

PR #222, opened 2025-03-17
Support RPMs that require a subscription

querti:
RPMs requiring a subscription can be accessed by specifying a cert+key generated by the subscription manager. There is an existing Konflux functionality that allows users to build images with RPMs that need a subscription[1]. Users need to add a secret to Konflux with the activation key that's used to generate the cert+key. MintMaker can use the same secret for its own purposes.

The general workflow is:
1. MintMaker controller searches for the activation key secret in user namespace
2. If found, it generates the same secret for the PipelineRun
3. PipelineRun uses the secret to generate cert+key
4. Cert+key are placed in a predictable location set to ENV variables
5. ENV variables propagate to the repo file, allowing rpm-lockfile-prototype to utilize the cert+key

The generation of cert+key requires root user, so this step is run as root. Other steps are still run as regular users and even the root container is not elevated, so there shouldn't be any major security concerns. Anyuid SCC was added tp tje service account cluster role, sp that it would have the permission to run containers as root. Specifying SeccompProfile in the task had to be removed, because it's seemingly disallowed with the anyuid SCC and the PipelineRun would fail to run. It shouldn't make a difference since the default profile is used anyway.

RPM repo file set sslclientkey and sslclientcert parameters to locations of the aforementioned cert+key[2]. Since the processing happens in a MintMaker pod, users cannot know what paths to set these parameters to. Thus can be overcome by setting them to ENV variables $SSL_CLIENT_KEY and $SSL_CLIENT_CERT. MintMaker itself sets those variables to valid paths, allowing the authentication to happen. THe prefix DNF_VAR_ is stripped by dnf. To take advantage of this functionality, users will need to be told to change the paths in their repo files to these variables.

[1] https://konflux.pages.redhat.com/docs/users/how-tos/configuring/activation-keys-subscription.html
[2] https://dnf.readthedocs.io/en/latest/conf_ref.html#repo-variables

Refers to CLOUDDST-26090

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -226,7 +226,6 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(comp component.GitCo
 	resources = append(resources, renovateSecret)
 
 	// Create Secret for RPM activation key to access RPMs that require subscription
-	//
 	activationKey, org, rpmKeyErr := comp.GetRPMActivationKey(r.Client, ctx)
 	var rpmSecret *corev1.Secret = nil
 	if rpmKeyErr != nil {
--- a/internal/pkg/component/base/base.go
+++ b/internal/pkg/component/base/base.go
@@ -23,8 +23,11 @@ import (
 	"strings"
 	"sync"
 
+	bslices "github.com/konflux-ci/mintmaker/internal/pkg/slices"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logger "sigs.k8s.io/controller-runtime/pkg/log"
 )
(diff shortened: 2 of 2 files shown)


staticf0x (dismissed):
LGTM, but does Konflux let us run something as root without any explicit config in infra-deployments?

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -109,37 +111,52 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+											},
+										},
+										{
+											Name:  "prepare-rpm-cert",
+											Image: "registry.access.redhat.com/ubi9",
+											Script: "[ ! -f \"/etc/renovate/secret/rpm-activationkey\" ] && echo 'RPM secret not found. Exiting.' && exit 0;" +
+												"echo 'Generating RPM certificate and copying it to shared workspace';" +
+												"KEY_NAME=$(cat /etc/renovate/secret/rpm-activationkey);" +
+												"ORG_ID=$(cat /etc/renovate/secret/rpm-org);" +
+												"subscription-manager register --activationkey=\"$KEY_NAME\" --org=\"$ORG_ID\";" +
```
qixiang, line 152: We should probably unregister the system from subscription after renovate runs, otherwise there will be one system added to the subscription each time when renovate is executed, which can result in a large amount of registered systems in subscription.

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -109,37 +111,52 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+												"mkdir -p /workspace/shared-data/rpm-certs;" +
+												"cp /etc/pki/entitlement/*-key.pem /workspace/shared-data/rpm-certs/key.pem;" +
```
qixiang, line 154: Do we need to copy the certificates under `/etc/pki/consumer/`?

qixiang (commented):
As Riley mentioned, I'm also worry about the privilege issue

internal/pkg/tekton/pipeline_run_builder.go
qixiang: iirc, the cert under `/etc/pki/consumer` is used to identify the system, but I didn't test what will happen if we only copy the entitlement certs.

internal/pkg/tekton/pipeline_run_builder.go
qixiang: I think it should be fine to leave it as is for now, until someone raises a concern or complaint about it :)

qixiang (commented):
I'm wondering if we can avoid running the cert prepare container without root user, the reason is I have plan to create a service account for each PLR dynamically, which doesn't work with this approach.

With one service account for each PipelineRun, we can benefit from Tekton's capability of mounting the image pull secrets automatically: https://tekton.dev/docs/pipelines/auth/#overview

Is it possible to mount an emptyDir to the cert prepare container with path of /etc/pki/entitlement, then mount to the renovate container with same path? then we don't need to copy the certs.

qixiang (approved):
LGTM

PR #223, opened 2025-03-18
Remove semanticCommits from packageRules

staticf0x:
This option doesn't do anything unless it's on top level. The default is `auto` ([docs](https://docs.renovatebot.com/configuration-options/#semanticcommits)), which detects, whether semantic commits are used in the project using [conventional-commits-detector](https://www.npmjs.com/package/conventional-commits-detector).

Renovate uses the last 20 commit messages to detect that. Up until now, if it looked like this option worked, it didn't. It just detected semantic commits automatically. Until we found a repo where it didn't work :smile: . So if it doesn't detect correctly, users need to set it explicitly to either `enabled` or `disabled`.

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -93,7 +93,6 @@
           "commitMessageTopic": "{{{groupName}}}"
         },
         "commitMessageTopic": "Konflux references",
-        "semanticCommits": "enabled",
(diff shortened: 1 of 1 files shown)


tonisbones (approved):
Good find, Riley!

nvtnlucie (approved):

PR #228, opened 2025-03-25
feat: set PipelineRun timeout to 1 hour

qixiang:
Set the PipelineRun timeout to 1 hour instead of relying on the default value from the Tekton Pipelines controller. This duration should provide sufficient time for Renovate jobs while preventing excessive runtime if something goes wrong.


staticf0x:
Is the timeout counted since the PLR starts or since it's created?

qixiang:
> Is the timeout counted since the PLR starts or since it's created?

Since the PLR starts, if a PipelineRun has pipeline timeout of 1h, but it's in pending for 2hours, it won't timeout.

PR #235, opened 2025-03-28
Remove timestamp field from component struct

qixiang:
The timestamp was previously used for implementing the refreshing of GitHub App installation cache. Since the cache is now refreshed based on time, we no longer need to store timestamp information in the component struct.

Additionally, changed the timestamp in PipelineRun name from Unix timestamp format to MMDDhhmm, making it more human-readable.

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -19,16 +19,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
-	"strconv"
 	"time"
 
-	appstudiov1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
-	mmv1alpha1 "github.com/konflux-ci/mintmaker/api/v1alpha1"
-	"github.com/konflux-ci/mintmaker/internal/pkg/component"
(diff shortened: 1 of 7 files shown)


internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -145,8 +147,8 @@ func (r *DependencyUpdateCheckReconciler) createMergedPullSecret(ctx context.Con
 		return nil, err
 	}
 
-	timestamp := time.Now().Unix()
-	name := fmt.Sprintf("renovate-image-pull-secrets-%d-%s", timestamp, utils.RandomString(5))
+	timestamp := time.Now().UTC().Format("01021504") // in MMDDhhmm format
```
staticf0x, line 150: Can we do something like
```go
const MMDDhhmm = "01021504"
timestamp := time.Now().UTC().Format(MMDDhhmm)
```

I didn't know Go has such cryptic datetime formatting

qixiang: This is indeed quite weird - who knows why Go's inventors decided to do it this way? You can get more info about this reference date from this link:
https://stackoverflow.com/questions/45160822/

I improved the comments to include more info about this formatting string.

staticf0x (approved):

PR #260, opened 2025-05-05
Enable prometheus SM

nvtnlucie:
The final changes to enable metric export via http which I tested on the local cluster. Should be merged together with https://github.com/redhat-appstudio/infra-deployments/pull/6296

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -63,7 +63,7 @@ func main() {
 	var secureMetrics bool
 	var enableHTTP2 bool
 	var tlsOpts []func(*tls.Config)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metrics endpoint binds to. "+
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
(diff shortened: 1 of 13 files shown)


config/default/metrics_service.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     app.kubernetes.io/name: mintmaker
     app.kubernetes.io/managed-by: kustomize
   name: controller-manager-metrics-service
-  namespace: system
```
qixiang, line 9: In our deployment, namespace is set in kustomization.yaml files, so it will be changed by kustomize. But it doesn't hurt to change this.

cmd/manager/main.go
```diff
@@ -63,7 +63,7 @@ func main() {
 	var secureMetrics bool
 	var enableHTTP2 bool
 	var tlsOpts []func(*tls.Config)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metrics endpoint binds to. "+
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
```
qixiang, line 66: what's the purpose of changing this default value?

qixiang (commented):
Do you know if we're allowed to use http in production? I think it's fine to test in staging with http.

cmd/manager/main.go
nvtnlucie: It was set to this value in the beginning (before the metric epic), so I decided it would look nicer if i kept it. The real value is set by kustomize anyway.

FernandesMF (approved):

PR #277, opened 2025-05-19
Increase cache times of repo cache and introduce config.js

querti:
In order to improve performance and reduce the rate limits, we should increase the package cache expiration. The duration is set to 4 hours, which represents one run of MintMaker. I think that having a slightly stale data is acceptable, since MintMaker runs only every 4 hours and only updates dependencies once a week for most users.

Self hosted config options can only be written in the config.js file and will be ignored in renovate.json. To resolve this, transform the mounted global config file to config.js. A new file, self_hosted.json, which holds only the self-hosted configuration options was created. Options that are already in use were transferred from env variables to this file as well. The file is combined with renovate.json and mounted to renovate as config.js.

During testing, some cache sets didn't respect the new values. This is due to how renovate was written, where the cache scaffolding is sometimes not used and so the configuration is ignored. Renovate's code needs to be modified to be able to tweak these values.


internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -188,14 +188,15 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(name string, comp co
 	if err != nil {
 		return nil, err
 	}
+	renovateJsConfig := "module.exports = " + renovateConfig
```
staticf0x, line 191: Is this necessary? Renovate can read plain JSON, or is that different for the self_hosted?

staticf0x (approved):

internal/controller/dependencyupdatecheck_controller.go
staticf0x: thanks I was a bit confused in the morning :sweat_smile:

PR #279, opened 2025-05-30
Disable repository cache

querti:
It looks like repository cache doesn't save RPM lockfiles properly, causing issues with this manager. Disable repo cache until the bug is fixed.


qixiang (approved):
:+1:

staticf0x (changes requested):
Please wait before merging, I don't think this is the problem

staticf0x (approved):

PR #281, opened 2025-06-05
Move go cache to a writable directory

querti:
"go get" used by renovate sometimes initializes a cache, which causes an error "failed to initialize build cache at /home/renovate/.cache/go-build: mkdir /home/renovate/.cache/go-build/00: permission denied". This causes go updates to behave incorrectly, for example only update go.mod file as seen in [1]. It can be fixed by moving the cache to a writable location /tmp by setting the GOCACHE env variable.

NOTE: This fix hasn't been tested because the error it's difficult to replicate locally. But this change should be harmless and we can observe if the errors have disappeared in the Splunk logs.

[1] https://github.com/openshift/sandboxed-containers-operator/pull/624/files

Refers to CWFHEALTH-4164


qixiang (dismissed):
:+1:

staticf0x (dismissed):

qixiang (approved):
LGTM

PR #294, opened 2025-06-26
Use migrate subcommand

tkdchen:
STONEBLD-3551

pipeline-migration-tool has subcommands now. migrate is the one handling migrations.

Requires pipeline-migration-tool v0.3.0


tkdchen:
Requires https://github.com/konflux-ci/mintmaker-renovate-image/pull/136

PR #295, opened 2025-06-26
Update pipeline-migration-tool regex match

querti:
It seems that renovate now matches the final command instead of the raw command, which breaks the regex match. Loosen the regex somewhat to resolve this issue.

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -401,7 +401,7 @@
   "forkProcessing": "enabled",
   "allowedCommands": [
     "^rpm-lockfile-prototype rpms.in.yaml$",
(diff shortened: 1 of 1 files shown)


staticf0x:
This is a bit problematic as it can be used for arbitrary command execution, see this config:

```
{
    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
    "tekton": {
        "postUpgradeTasks": {
            "commands": [
                "pipeline-migration-tool -u '[{\"depName\": \"'; echo 'hacked' >> .tekton/mintmaker-test-pull-request.yaml;#\", \"currentValue\": \"0.1\", \"currentDigest\": \"sha256:e6beb161ed59d7be26317da03e172137b31b26648d3e139558e9a457bc56caff\", \"newValue\": \"0.2\", \"newDigest\": \"sha256:517a51e260c0b59654a9d7b842e1ab07d76bce15ca7ce9c8fd2489a19be6463d\", \"packageFile\": \".tekton/mintmaker-test-push.yaml\", \"parentDir\": \".tekton\", \"depTypes\": [\"tekton-bundle\"]}]'"
            ]
        }
    }
}
```

PR #302, opened 2025-07-02
Added Config for MintMaker controller

HozifaWasfy:
- Created a ConfigMap-based configuration system (internal/pkg/config/config.go) that reads settings from K8s configMap with automatic fallback to sensible defaults if the ConfigMap doesn't exist or has invalid values.
- Replaced all hardcoded values across the codebase .


qixiang:
@HozifaWasfy could you squash the commits?

PR #303, opened 2025-07-03
Include PipelineRun completion timestamps in controller logs

KristianTkacik:
This MR modifies the PipelineRun controller to log the completion timestamps of finished PipelineRuns.

### DIFF
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


qixiang:
When a cluster contains a large number of PipelineRuns (in our case, one cluster may have around 2,000 PipelineRuns created by mintmaker in a short period), updating PipelineRuns can be an expensive operation. Therefore, introducing a finalizer to implement this functionality could impact performance.

Since we already monitor PipelineRun state changes in our PipelineRun controller, you can directly record timing information when relevant state changes occur, without needing to use finalizer for this purpose.

For example (pseudocode not tested):
```
func (r *PipelineRunReconciler) SetupWithManager(mgr ctrl.Manager) error {
    return ctrl.NewControllerManagedBy(mgr).
        For(&tektonv1.PipelineRun{}).
        WithEventFilter(predicate.Funcs{
            CreateFunc:
             ...
            },
            DeleteFunc: func(deleteEvent event.DeleteEvent) bool { return false },
            UpdateFunc: func(updateEvent event.UpdateEvent) bool {
                if oldPipelineRun, ok := updateEvent.ObjectOld.(*tektonv1.PipelineRun); ok {
                    if newPipelineRun, ok := updateEvent.ObjectNew.(*tektonv1.PipelineRun); ok {
                        // TODO: log when a PipelineRun finishes, there might be multiple states
                        // should be covered: succeeded, failed, canceled and etc, needs some testing.
                        if oldPipelineRun.Status.GetCondition(apis.ConditionSucceeded).IsUnknown() && 
                           !newPipelineRun.Status.GetCondition(apis.ConditionSucceeded).IsUnknown() {
                            if newPipelineRun.Status.CompletionTime != nil {
                                completionTime := newPipelineRun.Status.CompletionTime.Format(time.RFC3339)
                                // include pipelinerun name and corresponding state in log
                                log.Info(fmt.Sprintf("PipelineRun finished at %s", completionTime))
                            }
                        }
                        ...
                    }
                }
                return false
            },
            GenericFunc: func(genericEvent event.GenericEvent) bool { return false },
        }).
        Complete(r)
```
It's not ideal to mix such code in filtering, but much more efficient than do that in Reconcile (you'll need to get the resource and check the state again).

PR #304, opened 2025-07-04
Start PipelineRuns based on creation order

qixiang:
In the original implementation, we started GitHub PipelineRuns first in the PipelineRun controller. This was necessary because if GitHub PipelineRuns waited too long (e.g., 1 hour), the token used during creation would expire.

Since we've now added GitHub repository token renewal functionality, this original design is no longer necessary. We can simplify the implementation by starting PipelineRuns based on their creation order.

JIRA: CWFHEALTH-4081


staticf0x (approved):
LGTM

PR #308, opened 2025-07-09
fix: log completion timestamps only for MintMaker PipelineRuns

KristianTkacik:


internal/controller/pipelinerun_controller.go
```diff
@@ -228,7 +228,8 @@ func (r *PipelineRunReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					if newPipelineRun, ok := updateEvent.ObjectNew.(*tektonv1.PipelineRun); ok {
 						if oldPipelineRun.Status.GetCondition(apis.ConditionSucceeded).IsUnknown() &&
 							!newPipelineRun.Status.GetCondition(apis.ConditionSucceeded).IsUnknown() {
-							if newPipelineRun.Status.CompletionTime != nil {
+							if newPipelineRun.Namespace == MintMakerNamespaceName &&
+								newPipelineRun.Status.CompletionTime != nil {
```
qixiang, line 232: This check can be moved earlier for better efficiency, reducing nesting levels and improving readability. 
```
if e.Object.GetNamespace() != MintMakerNamespaceName {
	return false
}
```
I don't remember why we used more specific variable names (`createEvent`, `updateEvent` and etc), but we can just change them to a more generic variable name `e`.

Furthermore, this check can be added to other event handlers to avoid having to make this check in the `Reconcile` function:
```
	// Only process events in the MintMaker namespace
	if req.Namespace != MintMakerNamespaceName {
		return ctrl.Result{}, nil
	}
```

qixiang (approved):
Nice, thanks!

PR #316, opened 2025-07-18
Log pipelinerun cancellations

qixiang:
This change adds an info-level log message when a pipelinerun is successfully cancelled, improving observability and making it easier to trace the pipelinerun history.


KristianTkacik (approved):
LGTM

PR #318, opened 2025-07-21
Pass upgrades to migration tool via file

tkdchen:
STONEBLD-3559

This change requires Renovate>=41.1.0 which introduces dataFileTemplate option.


staticf0x (approved):
Tested on our E2E repo with the new Renovate version and pipeline-migration-tool 0.4.0, everything seems to work like a charm. Thank you!

PR #322, opened 2025-07-25
Use user-provided image registry secrets

KristianTkacik:
Previously, MintMaker used global pull secrets for registry.redhat.io and registry.stage.redhat.io. These secrets were applied to all PipelineRuns, regardless of need. This approach created a security issue by allowing any user to access these secure registries, even without proper permissions.

This commit updates the DependencyUpdateCheck controller to use user-provided image registry secrets instead. For images onboarded to Konflux, these secrets are expected to be provided and linked to a component-specific service account in the user namespace called build-pipeline-$COMPONENT_NAME.

JIRA: CWFHEALTH-4113

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -83,96 +84,90 @@ func (r *DependencyUpdateCheckReconciler) getCAConfigMap(ctx context.Context) (*
...
-	if len(secretList.Items) == 0 {
-		// No secrets to merge
-		return nil, nil
+	linkedSecrets := make(map[string]struct{})
+	for _, secret := range serviceAccount.Secrets {
+		linkedSecrets[secret.Name] = struct{}{}
 	}
 
-	log.Info(fmt.Sprintf("Found %d secrets to merge", len(secretList.Items)))
+	var allSecrets corev1.SecretList
```
qixiang, line 110: Since we already have secrets linked to this service account, we don't need to retrieve all secrets in the namespace and check them individually. Instead, we can directly get and check each secret one by one. While this might result in a few more API calls, it's worth considering that some namespaces could have numerous components, for example, OpenStack team have hundreds of components with potentially over 1,000 secrets. If we process each component by listing and checking all secrets every time, this would significantly impact performance.

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -83,96 +84,90 @@ func (r *DependencyUpdateCheckReconciler) getCAConfigMap(ctx context.Context) (*
...
-			var dockerConfig map[string]interface{}
-			if err := json.Unmarshal(data, &dockerConfig); err != nil {
-				return nil, err
-			}
+		var secretDataKey string
+		switch secret.Type {
+		case corev1.SecretTypeDockerConfigJson:
+			secretDataKey = corev1.DockerConfigJsonKey
+		case corev1.SecretTypeDockercfg:
+			secretDataKey = corev1.DockerConfigKey
```
qixiang, line 127: Seems the `dockercfg` format is not compatible with the `.docker/config.json` format, I think we can ignore such secrets. The other reason is for each SA, there are some secrets created automatically, one is the internal used image pull secret, which is for the in-cluster registry urls, it's also in `dockercfg` format, we don't need to include secrets for such registries.

Another secret created by konflux (linked to the SA) is for pushing images, its type is `kubernetes.io/dockerconfigjson`, but it has a label of `appstudio.redhat.com/internal=true`, I think we can also skip such secrets.

internal/controller/dependencyupdatecheck_controller.go
```diff
@@ -326,15 +331,17 @@ func (r *DependencyUpdateCheckReconciler) createPipelineRun(name string, comp co
...
+	if renovateSecret.Data != nil {
+		if _, exists := renovateSecret.Data[corev1.DockerConfigJsonKey]; exists {
+			secretItems := []corev1.KeyToPath{
+				{
+					Key:  corev1.DockerConfigJsonKey,
+					Path: "config.json",
+				},
+			}
+			secretOpts := tekton.NewMountOptions().WithTaskName("build").WithStepNames([]string{"renovate"}).WithReadOnly(true)
+			builder.WithSecret(renovateSecret.ObjectMeta.Name, "/home/renovate/.docker", secretItems, secretOpts)
```
qixiang, line 343: This will generate another volume with name of the secret, but we already have one volume created with the same name (for mounting the renovate token) before this, two volumes with same name is not allowed in k8s/openshift.

It would be ideal to create one volume, but mount two keys to different locations, however that requires refactoring the pipeline builder. An easier solution is creating two volumes with same secret, but using different names, for example in `WithSecret`, we can modify it to:
```
-			volumeName := fmt.Sprintf("secret-%s", name)
+			// Generate unique volume name using random string to avoid conflicts
+			// when the same secret is mounted multiple times
+			volumeName := fmt.Sprintf("secret-%s-%s", name, utils.RandomString(8))
```

qixiang (changes requested):
We need to handle cases where components don't have any user provided image pull secrets. Currently in `TransformHostRules`, we're checking if the `registrySecret` exists, but since the secret will always be created now (we're using the same secret contains renovate token), we should instead verify the presence of specific keys within the secret for image pull credentials. Without this check, components without registry authentication will encounter unmarshal errors (though in current implementation, we might not see these errors due to default secrets created for each component's build service account).

A better approach would be to add a method that specifically retrieves image pull secrets for each component. We could then call this method both when creating the secret and when generating the hostRules (better to cache it). You can continue with the current implementation and improve this later.

internal/controller/dependencyupdatecheck_controller.go
KristianTkacik: Good point! I changed the method to only retrieve the linked secrets instead of all secrets in the namespace.

internal/controller/dependencyupdatecheck_controller.go
KristianTkacik: Thank you for the suggestion! I changed the implementation to ignore also `dockercfg` and internal Secrets.

internal/controller/dependencyupdatecheck_controller.go
KristianTkacik: Nice catch, thank you for pointing this out! I applied your suggestion to the `WithSecret` function.

qixiang (approved):
The queue system has been deployed to all clusters, feel free to merge this.

PR #327, opened 2025-07-31
Remove pending status and concurrent limit for PipelineRuns

qixiang:
The upcoming queue system in Konflux cannot properly handle PipelineRuns with a pending status. Therefore, we need to update the controller to create PipelineRuns without setting them to pending status initially.

Previously, setting the pending status was used to control the maximum number of concurrently running PipelineRuns. However, with the introduction of the queue system, we will not be able to control that anymore. Consequently, the related code (which moves PipelineRuns from pending to running status) also needs to be removed.

### DIFF
--- a/internal/controller/dependencyupdatecheck_controller.go
+++ b/internal/controller/dependencyupdatecheck_controller.go
@@ -491,7 +491,7 @@ func (r *DependencyUpdateCheckReconciler) Reconcile(ctx context.Context, req ctr
 			processedComponents = append(processedComponents, key)
 		}
 
-		log.Info(fmt.Sprintf("creating pending PipelineRun for %s", key))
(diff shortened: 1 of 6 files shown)


qixiang:
Notes:
1. Actually we can remove the PipelineRun controller without affecting functionality. I'm keeping the PipelineRun controller for two reasons: first, to retain relevant logging that helps us debug issues; second, because we may need to implement other features in the PipelineRun controller later (e.g. the issue dashboard related features?).

2. Please do not merge this PR until the queue system is officially deployed to production or until we know the exact deployment timeline. The existing concurrency control functionality is still needed before the queue system goes live. If we merge this PR without a clear timeline for the queue system deployment, it will block the release of our subsequent commits.

PR #332, opened 2025-08-12
Add an option to enable pprof in manager

qixiang:


qixiang:
Initially, I suspected that the memory issue in the p02 cluster was due to fetching all components first and then processing them sequentially. However, after testing locally with 5000+ components, it appears this isn't the cause; the memory allocation for storing components looks fine. To further investigate, I'm adding this option for enabling profiling. This will allow me to debug the issue using profiling data from our production clusters.

PR #336, opened 2025-08-14
Add namespace filtering back to controller event filters

qixiang:
The builder-level predicates added in commit 6ae8669 should filter at informer level, but seems events from other namespaces are still reaching controller event filters.

From observation, it seems that only logging in event filters is affected, and the events are not actually processed by Reconcile. Until the exact impact is confirmed, restore the namespace filter to ensure correctness.


FernandesMF (approved):
Makes sense and looks good to me!

PR #343, opened 2025-08-25
Add auto assign action

staticf0x:
Trying out the GitHub action that will assign reviewers to new PRs. I removed `CODEOWNERS` because it causes everyone from the list to be a reviewer and could interfere with this action. The action is pinned to a specific commit instead of a tag to prevent hijacks that already happened recently on GitHub.


qixiang (approved):
LGTM

PR #350, opened 2025-08-29
Migrate users from redhat-appstudio-tekton-catalog to konflux-ci/tekton-catalog

chmeliik:
### Problem

Around June 2024, we started releasing Tekton tasks to the quay.io/konflux-ci/tekton-catalog namespace.

But we did not stop releasing them to quay.io/redhat-appstudio-tekton-catalog, because all existing Konflux users at the time would have the redhat-appstudio-tekton-catalog references in their .tekton/ yaml files.

We did not know of any good way to migrate the users over to the new konflux-ci/tekton-catalog references, so we just kept releasing to both places.

Now, with the decentralization of build-definitions in progress, we are starting to release Tekton tasks via Konflux, which will make it hard (if not impossible) to keep releasing to the old location.

---

### Solution

Configure replacementName [1] for all tasks that exist both in redhat-appstudio-tekton-catalog and in konflux-ci/tekton-catalog. This will cause Mintmaker to send out PRs that automatically migrate users away from redhat-appstudio-tekton-catalog.

Note: I'm aware that replacementNameTemplate [2] exists, and that it would make this PR much, much shorter. Unfortunately, Renovate seems to have a bug that makes replacementNameTemplate unusable in this case: https://github.com/renovatebot/renovate/discussions/37744

---

Make sure to group replacement updates separately, so that Mintmaker ends up sending two PRs, one that groups the regular digest/version updates and one that groups the replacement updates. 

[1]: https://docs.renovatebot.com/configuration-options/#replacementname
[2]: https://docs.renovatebot.com/configuration-options/#replacementnametemplate

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -101,6 +101,396 @@
         "prBodyTemplate": "{{{header}}}{{{table}}}{{{notes}}}{{{changelogs}}}{{{configDescription}}}{{{controls}}}{{{footer}}}",
         "recreateWhen": "always",
         "rebaseWhen": "behind-base-branch"
+      },
+      {
+        "matchPackageNames": [
+          "quay.io/redhat-appstudio-tekton-catalog/task-acs-deploy-check"
+        ],
+        "replacementName": "quay.io/konflux-ci/tekton-catalog/task-acs-deploy-check"
+      },
+      {
+        "matchPackageNames": [
+          "quay.io/redhat-appstudio-tekton-catalog/task-acs-image-check"
+        ],
+        "replacementName": "quay.io/konflux-ci/tekton-catalog/task-acs-image-check"
+      },
+      {
+        "matchPackageNames": [
+          "quay.io/redhat-appstudio-tekton-catalog/task-acs-image-scan"
+        ],
+        "replacementName": "quay.io/konflux-ci/tekton-catalog/task-acs-image-scan"
+      },
+      {
+        "matchPackageNames": [
+          "quay.io/redhat-appstudio-tekton-catalog/task-apply-tags"
+        ],
+        "replacementName": "quay.io/konflux-ci/tekton-catalog/task-apply-tags"
+      },
+      {
+        "matchPackageNames": [
+          "quay.io/redhat-appstudio-tekton-catalog/task-buildah"
+        ],
+        "replacementName": "quay.io/konflux-ci/tekton-catalog/task-buildah"
+      },
+      {
+        "matchPackageNames": [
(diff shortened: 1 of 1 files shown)


chmeliik:
Tested this in https://github.com/chmeliik/integration-service/pull/12 and https://github.com/chmeliik/integration-service/pull/16, but with the upstream Renovate. I have onboarded the repo to Mintmaker and am currently waiting for the Mintmaker update, to verify that the configuration will work the same way.

qixiang:
@chmeliik since this is part of the Tekton manager, usually when the Tekton manager is executed, there are some tasks have updates, and pipeline migration tool is also executed against each task. I have two questions:
1. what happens if at least one task has updates?
2. does it work well with pipeline migration tool?

chmeliik:
In https://github.com/chmeliik/integration-service/pull/18, I think the branch name is as expected:
image

If I understand correctly, Renovate does a deep merge on the individual `packageRules` objects. So the packageRules that matches replacement-type updates inherits `group.branchTopic` from the rule that precedes it.

But I can add `group.branchTopic` explicitly if that reduces confusion

qixiang:
> In chmeliik/integration-service#18, I think the branch name is as expected

By default, with upstream renovate, your repository is onboarded with all branches, in this case, renovate appends the branch name to source branch names automatically. But mintmaker executes renovate against single branch of your repository, so we need to add the branch name explicitly.

> If I understand correctly, Renovate does a deep merge on the individual `packageRules` objects. So the packageRule object that matches replacement-type updates inherits `group.branchTopic` from the rule that precedes it.
> But I can add `group.branchTopic` explicitly if that reduces confusion

I'm uncertain about this, you may need to test it to confirm whether that can be merged and applied to your new "group".

chmeliik:
> > In [chmeliik/integration-service#18](https://github.com/chmeliik/integration-service/pull/18), I think the branch name is as expected
> 
> By default, with upstream renovate, your repository is onboarded with all branches, in this case, renovate appends the branch name to source branch names automatically. But mintmaker executes renovate against single branch of your repository, so we need to add the branch name explicitly.
> 
> > If I understand correctly, Renovate does a deep merge on the individual `packageRules` objects. So the packageRule object that matches replacement-type updates inherits `group.branchTopic` from the rule that precedes it.
> > But I can add `group.branchTopic` explicitly if that reduces confusion
> 
> I'm uncertain about this, you may need to test it to confirm whether that can be merged and applied to your new "group".

https://github.com/chmeliik/integration-service/pull/18 is from Mintmaker :)

staticf0x:
Let me set up a E2E test from your repos, it will be very valuable for the future as well

PR #352, opened 2025-09-01
Update go packages

qixiang:
- Upgrade Go version to 1.25.0
- Update dependencies
- Important: Pin knative.dev/pkg to avoid breaking changes where knative.dev/pkg/metrics is no longer available after Knative's migration to OpenTelemetry


qixiang:
> > Important: Pin knative.dev/pkg to avoid breaking changes where knative.dev/pkg/metrics is no longer available after Knative's migration to OpenTelemetry
> 
> Should we switch too? Is it even possible in Konflux right now?

It's not directly used by us, it's used by tekton and tekton needs to migrate, then we can remove this version pinning.

PR #368, opened 2025-09-19
Re-enable pre-commit and git-submodules

staticf0x:
This was accidentally removed during #364 

`pre-commit` is disabled by default, so we need the explicit `enabled: true`

Edit: same with `git-submodules`

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -566,6 +566,12 @@
     "rebaseWhen": "behind-base-branch",
     "branchTopic": "lock-file-maintenance"
   },
+  "git-submodules": {
+    "enabled": true
+  },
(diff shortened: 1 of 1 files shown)


config/renovate/renovate.json
```diff
@@ -566,6 +566,12 @@
     "rebaseWhen": "behind-base-branch",
     "branchTopic": "lock-file-maintenance"
   },
+  "git-submodules": {
+    "enabled": true
+  },
+  "pre-commit": {
+    "enabled": true
+  },
```
qixiang, line 574: oh, I thought this `enabled` setting was redundant (at least for `dockerfile` manager), so when I removed the schedules, I deleted them as well.

PR #381, opened 2025-10-03
feat: Send PipelineRun successes and failures to KITE API

adelaon:
Add functions for sending requests to defined webhooks. Extend pipelinerun_controller Reconcile function to watch for pipelinerun failure and success and send corresponding requests.

{"error":"Missing namespace"} bug resolved - add namespace as query param

CWFHEALTH-4404

### DIFF
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/apis"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -29,8 +30,10 @@ import (
 
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 
+	appstudiov1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
 	"github.com/konflux-ci/mintmaker/internal/pkg/config"
 	. "github.com/konflux-ci/mintmaker/internal/pkg/constant"
+	"github.com/konflux-ci/mintmaker/internal/pkg/kite"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


internal/pkg/kite/client.go
```diff
@@ -89,3 +110,31 @@ func (c *Client) GetVersion(ctx context.Context) (string, error) {
 
 	return verStr, nil
 }
+
+// SendPipelineFailure sends a pipeline failure event to KITE webhook
+func (c *Client) SendPipelineFailure(ctx context.Context, payload PipelineFailurePayload) error {
+	url := fmt.Sprintf("%s/api/v1/webhooks/pipeline-failure?namespace=%s", c.baseURL, payload.Namespace)
+
+	jsonPayload, _ := json.Marshal(payload)
+	fmt.Printf("Sending failure request to %s containing \n %s", url, string(jsonPayload))
```
staticf0x, line 119: Wouldn't printf just print to stdout instead of structured logging? Would this be correctly forwarded to Splunk?

internal/controller/pipelinerun_controller.go
```diff
@@ -41,15 +43,83 @@ var (
...
+
+	// Get the PipelineRun object
+	var pipelineRun tektonv1.PipelineRun
+	if err := r.Client.Get(ctx, req.NamespacedName, &pipelineRun); err != nil {
+		// PipelineRun was deleted, nothing to do
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Only process completed PipelineRuns
+	if !pipelineRun.IsDone() {
```
staticf0x, line 64: Instead of checking the PLR status here, could we add a callback to the checks in `SetupWithManager`?

internal/controller/pipelinerun_controller.go
```diff
@@ -41,15 +43,83 @@ var (
...
+		return ctrl.Result{}, nil
+	}
+
+	// Check if the PipelineRun failed
+	condition := pipelineRun.Status.GetCondition(apis.ConditionSucceeded)
+	if condition != nil && !condition.IsTrue() {
```
staticf0x, line 70: I'd split this in a way that if `condition` is nil, use early return and then do
```go
if condition.IsTrue() {}
else {}
```

to make it more readable.

internal/controller/pipelinerun_controller.go
```diff
@@ -41,15 +43,83 @@ var (
...
+		// send failure webhook to KITE
+		if r.KiteClient != nil {
+			if err := r.sendFailureWebhook(ctx, &pipelineRun); err != nil {
+				fmt.Printf("\nfailed to send pipeline failure webhook for pipelineRun %s: %s\n", pipelineRun.Name, err)
```
staticf0x, line 74: Same as I mentioned in the other file, can we use the `log` object here?

FernandesMF:
I just saw @staticf0x's suggestions. Mine will have an overlap with hers.

PR #386, opened 2025-10-15
feat: Add another step to the pipeline (for analyzing logs and integration with KITE API)

adelaon:
Instead of simple reason (like "Failed"), send detailed information extracted from pod logs to the webhook payload.

Added package 'doctor' with log_reader.go with functions to:
- Locate failed TaskRun pods within a PipelineRun
- Stream and parse container logs from Kubernetes API
- Parse Renovate's JSON log format into structured Go types
- Extract and format ERROR (level 50), and FATAL (level 60) messages (the immediate cause of non-zero exit)
- Build failure summary for KITE webhook payload

Refactor code:
- Use single function to send requests to all webhooks.
- Make destination URL creation safer using the net/url and path packages.

[CWFHEALTH-4405](https://issues.redhat.com/browse/CWFHEALTH-4405)

Assisted-by: Cursor and Gemini


qixiang:
Another reason I'm considering running the log analyzer as a separate step within the PipelineRun is that, with the current implementation, enabling or disabling this functionality isn't very straightforward. While it's possible, the implementation is rather messy. By simply adding a new step, we could easily enable or disable this feature during deployment.

staticf0x:
@qixiang your approach sounds good, perhaps more maintainable long term, but we should also consider that all the KITE related code would need to be duplicated, because you need it to report PLR fails and that information is only available in the controller and then you would need it in the log-analyzer program again after parsing the logs to send the data to KITE. Of course this could be solved by a shared package and two different binaries (controller and log-analyzer). Just a thing to consider.

I'm also inclined to go with the separate step approach.

internal/pkg/tekton/pipeline_run_builder.go
```diff
@@ -211,6 +232,43 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+														FieldRef: &corev1.ObjectFieldSelector{
+															FieldPath: "metadata.name",
+														},
+													},
+												},
+												{
+													Name: "NAMESPACE",
+													ValueFrom: &corev1.EnvVarSource{
+														FieldRef: &corev1.ObjectFieldSelector{
+															FieldPath: "metadata.namespace",
```
qixiang: Which namespace should be specified in the Kite requests? Is it always "mintmaker", or should it be the namespace of the component?
If it's the latter, then you should use the `mintmaker.appstudio.redhat.com/namespace` label instead.

qixiang (commented):
Just a minor comment, the rest looks good to me.

PR #490, opened 2026-04-03
chore: enable gitlint for PR commit message linting

qixiang:
Add gitlint configuration and a GitHub Actions job to enforce commit message standards on pull requests. Enabled rules:
- Conventional commit prefixes (chore|docs|feat|fix|refactor|style|test|revert)
- Lowercase description after prefix (custom UC1 rule)
- Title max length: 72 characters
- Body max line length: 88 characters
- Empty line between title and body
- Bot commits ([bot] author) are excluded from linting

Assisted-by: Claude Opus 4.6


staticf0x:
Should we also add `.pre-commit-config.yaml`?

```yaml
-   repo: https://github.com/jorisroovers/gitlint
    rev:  v0.19.1
    hooks:
    -   id: gitlint
```

The only caveat with gitlint is that you need to run `pre-commit install --hook-type commit-msg` instead of just `pre-commit install`, source: https://jorisroovers.com/gitlint/latest/commit_hooks/#pre-commit

We could perhaps use pre-commit for other hooks as well.

qixiang:
> Should we also add `.pre-commit-config.yaml`?

I'd prefer not adding that. The CI job already catches non-conforming commit messages, so enforcement is covered. A local pre-commit hook would only give faster feedback, but isn't essential. I think it's better to leave it to individual developers to decide whether to install a local hook (and its dependencies like python), that keeps things more flexible.

staticf0x (approved):

PR #498, opened 2026-04-14
feat(renovate): prefix Mintmaker PR titles and commits with base branch

bharath-b-rh:
## Summary

Set a top-level `commitMessagePrefix` in the shared Mintmaker Renovate config so that pull request titles and commit messages include the target base branch (for example `[release-1.19] …`, `[main] …`) for easier identification.

## Change

Add `"commitMessagePrefix": "[{{baseBranch}}] "` to `config/renovate/renovate.json`.

## Notes

Repos that set their own `commitMessagePrefix` continue to override this preset.

## References

[Renovate commitMessagePrefix](https://docs.renovatebot.com/configuration-options/#commitmessageprefix)
[Renovate templates (baseBranch)](https://docs.renovatebot.com/templates/)

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -25,6 +25,7 @@
   },
   "additionalBranchPrefix": "{{baseBranch}}/",
   "branchPrefix": "konflux/mintmaker/",
+  "commitMessagePrefix": "[{{baseBranch}}] ",
   "enabledManagers": [
     "github-actions",
     "tekton",


staticf0x:
Hi, where is this coming from? I think this should be up to users to configure, not centrally for all repos.

bharath-b-rh:
I just thought it would be good to have this additional detail on the PR, which will be helpful when we have multiple release branches and don't have a custom config. And if any projects have their own renovate config with this preset and it will be take the precedence.

staticf0x:
The problem is that this would also enable the prefix for users who have not configured their own and not all repos use a multibranch model, most actually only use `main` where this prefix would introduce more noise. I'd suggest keeping this configuration either in each repo where it is desired or through org inherited config.

PR #526, opened 2026-06-01
docs: add GitHub local controller testing agent skill

adelaon:
Document the end-to-end workflow for running MintMaker locally against minikube/kind with a GitHub-hosted test repo, including cluster bootstrap, GitHub App setup, Apple Silicon/Lima amd64 guidance, and smoke-test success criteria. Ship reusable Component and DUC templates under skills/local-controller-testing-workflow-github/assets/.


skills/local-controller-testing-workflow-github/SKILL.md
```diff
@@ -0,0 +1,419 @@
...
+The controller is **not** deployed into the cluster for this workflow.
+
+Test YAML templates live in `skills/local-controller-testing-workflow-github/assets/` — use those paths; do not rely on personal Desktop folders or other out-of-repo paths, unless instructed otherwise.
+
+## For AI agents — required before starting
+
+**Do not guess credentials or repository URLs.** Ask the user for:
+
+1. **GitHub App ID** — numeric ID from GitHub App settings (not the client ID).
+2. **GitHub App private key path** — absolute path to the `.pem` file on the developer's machine.
```
staticf0x, line 28: Will the private key be sent to the LLM?

skills/local-controller-testing-workflow-github/SKILL.md
```diff
@@ -0,0 +1,419 @@
...
+
+If `kubectl get nodes` fails with authentication errors, you are likely on a remote/OpenShift context. Switch to minikube/kind or start a local cluster.
+
+```bash
+kubectl config current-context
+kubectl config get-contexts -o name
+kubectl get nodes -o wide
+```
+
+### 2) Confirm a local cluster exists
```
KristianTkacik: nitpick: I think it might make more sense to have this as the first check

skills/local-controller-testing-workflow-github/SKILL.md
```diff
@@ -0,0 +1,419 @@
...
+# expect: one new row, REASON Running (or Unknown), name matching controller log
+```
+
+**Smoke test passed** when items 1–8 and the kubectl checks succeed. You do **not** need the PipelineRun to finish for day-to-day controller work.
+
+### Non-fatal messages (local testing only)
+
+These appear in **info** logs and do not block PipelineRun creation (rows 6–7 in the table above are expected locally):
+
+- `config file not found, using defaults` with `path":"/etc/mintmaker/config.json"` — no mounted operator config locally
```
KristianTkacik:
```suggestion
- `config file not found, using defaults` with `"path":"/etc/mintmaker/config.json"` — no mounted operator config locally
```

skills/local-controller-testing-workflow-github/SKILL.md
```diff
@@ -0,0 +1,419 @@
...
+```
+
+## Troubleshooting
+
+### Port 8081 already in use
+
+```bash
+lsof -i :8081
+kill -9 <pid>
+```
```
KristianTkacik, line 393:
```suggestion
```bash
lsof -i :8081
kill <pid>
# If the process does not shut down, force terminate it:
# kill -9 <pid>
```
Attempt to terminate it gracefully first.

KristianTkacik (commented):
@adelaon I only left a couple minor suggestions, but otherwise LGTM! Nice job!

skills/local-controller-testing-workflow-github/SKILL.md
```diff
@@ -0,0 +1,419 @@
...
+```bash
+minikube status
+kind get clusters
+```
+
+Only one of these needs to succeed.
+
+### 3) Check for stale test resources (important)
+
+A `DependencyUpdateCheck` processes **every** `Component` in namespaces listed in its spec. Leftover Components from earlier sessions cause extra PipelineRuns and memory pressure.
```
staticf0x: Wouldn't it be better to switch the logic, as in: keep whatever Components already exist, but specify which Components to run on in the DUC template?

skills/local-controller-testing-workflow-github/SKILL.md
adelaon: No, it should only ask for the path and then get only that. Then pass the string into the kubectl command. Unless a user attaches the `.pem` file directly in the chat or approves a shell command (like cat, type, head, or kubectl get secret with -o yaml|json). Also, if a user would put the `.pem` file in the workspace, then the LLM might get access to the secret. But by simply providing the path the LLM should not be able to read the file.

That said, just to be safe, I can add something like this into the skill:
```
- The private-key path is only for substituting into the documented kubectl create secret ... --from-file=... command.
- Never Read, @-mention, or shell-cat that file.
- Never dump the resulting Secret with -o yaml / -o json.
- If something goes wrong with the secrets, stop and ask the user to run the kubectl line themselves.
```

staticf0x: Thanks, I'd go for the safer route just to be sure

KristianTkacik (approved):

PR #532, opened 2026-06-08
Make MM resources readable to Konflux users

FernandesMF:
This commit enables any user authenticated in the Konflux cluster to read Konflux resources in the MintMaker namespace. The access is given via the 'konflux-viewer-user-actions' ClusterRole.

This will be important to enable the UI to view pipelineruns and their logs.

We are replicating the strategy from the rhtap-release tenant.

Assisted by: Claude

### Remarks
- the permissions of the 'konflux-viewer-user-actions' ClusterRole can be checked [here](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-rbac/production/base/konflux-viewer-user-actions.yaml)
- `make generate manifests` didn't produce any changes to the files involved in this PR

### DIFF
--- a/config/rbac/konflux_viewers.yaml
+++ b/config/rbac/konflux_viewers.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  labels:
+    app.kubernetes.io/name: mintmaker
+    app.kubernetes.io/managed-by: kustomize
+  name: mintmaker-tenant-konflux-viewer
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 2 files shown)


FernandesMF:
@adelaon yes, any user that authenticates to the Konflux cluster will be able to see all PLRs and their logs:
https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-rbac/production/base/konflux-viewer-user-actions.yaml#L39-L48

They will also be able to see pods and their logs:
https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-rbac/production/base/konflux-viewer-user-actions.yaml#L91-L100

This is the same strategy that the Release team adopts,

adelaon (approved):

PR #540, opened 2026-06-19
Implement log sanitization step

FernandesMF:
This commmit adds a new step in the MintMaker pipelinerun, to sanitize renovate-step logs.

In order to integrate MM with the Konflux UI, we will give read-permission to all logged users. Since the logs of all components lie in our namespace, all users will have access to the logs of all components.

If a credential were leaked in the logs, the wide access would make it severe. To prevent that, this commmit adds a step to sanitize the logs with leaktk, before they are persisted.

### Testing remarks

I tested manually that the script works as intended, and that the container that is created is able to run it.

### DIFF
--- a/internal/constant/constants.go
+++ b/internal/constant/constants.go
@@ -31,4 +31,7 @@ const (
 
 	RenovateImageEnvName    = "RENOVATE_IMAGE"
 	DefaultRenovateImageURL = "quay.io/konflux-ci/mintmaker-renovate-image:latest"
+
+	LeakTKImageEnvName    = "LEAKTK_IMAGE"
+	DefaultLeakTKImageURL = "quay.io/leaktk/leaktk:latest"
 )
--- a/internal/tekton/log_sanitizer.sh
+++ b/internal/tekton/log_sanitizer.sh
@@ -1,0 +1,62 @@
+#!/bin/sh
+LOG_FILE="${LOG_FILE:-/workspace/shared-data/renovate-logs.json}"
+FAIL_MSG='Sanitization step failed; the entire logs of this execution were removed for caution'
+fail_safe() {
+  echo "$FAIL_MSG"
+  echo "$FAIL_MSG" > "$LOG_FILE"
+  exit 0
+}
+if [ ! -f "$LOG_FILE" ]; then
+  echo 'Log file not found, skipping sanitization'
+  exit 0
+fi
+echo 'Scanning log file for leaked secrets...'
+SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>/dev/null)
+if [ $? -ne 0 ]; then
+  echo 'leaktk scan failed'
+  fail_safe
+fi
+if [ -z "$SCAN_OUTPUT" ]; then
+  echo 'No secrets detected or scanner produced no output'
+  exit 0
+fi
+SECRETS=$(echo "$SCAN_OUTPUT" | python3 -c "
+import sys, json
+for r in json.load(sys.stdin).get('results',[]):
+    s = r.get('secret','')
+    if s: print(s)
+")
+if [ $? -ne 0 ]; then
+  echo 'Failed to parse leaktk output'
+  fail_safe
+fi
+if [ -z "$SECRETS" ]; then
+  echo 'No secrets found in log file'
+  exit 0
+fi
+if ! cp "$LOG_FILE" "${LOG_FILE}.tmp"; then
+  echo 'Failed to create temporary log file'
+  fail_safe
+fi
+REDACT_FAILED=0
+while IFS= read -r secret; do
+  if [ -n "$secret" ]; then
+    # Prepend '\' to BRE metacharacters and the sed delimiter (|) so sed treats them as literal characters
+    ESCAPED=$(printf '%s\n' "$secret" | sed 's/[[\.*^$|\\]/\\&/g')
+    if ! sed -i "s|${ESCAPED}|**REDACTED**|g" "${LOG_FILE}.tmp"; then
+      REDACT_FAILED=1
+      break
+    fi
+  fi
+done <<EOF
+$SECRETS
+EOF
+if [ "$REDACT_FAILED" -eq 1 ]; then
+  echo 'Failed to redact secrets from log file'
+  fail_safe
+fi
+if ! mv "${LOG_FILE}.tmp" "$LOG_FILE"; then
+  echo 'Failed to replace log file with redacted version'
+  fail_safe
+fi
+echo 'Log sanitization complete'
--- a/internal/tekton/pipeline_run_builder.go
+++ b/internal/tekton/pipeline_run_builder.go
@@ -15,6 +15,7 @@
 package tekton
 
 import (
+	_ "embed"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -35,6 +36,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
+
+//go:embed log_sanitizer.sh
+var logSanitizerScript string
 
 type PipelineRunBuilder struct {
 	err         *multierror.Error
@@ -102,6 +106,10 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 	renovateImageURL := os.Getenv(mmconst.RenovateImageEnvName)
 	if renovateImageURL == "" {
 		renovateImageURL = mmconst.DefaultRenovateImageURL
+	}
+	leaktkImageURL := os.Getenv(mmconst.LeakTKImageEnvName)
+	if leaktkImageURL == "" {
+		leaktkImageURL = mmconst.DefaultLeakTKImageURL
 	}
 	return &PipelineRunBuilder{
 		pipelineRun: &tektonv1.PipelineRun{
@@ -236,6 +244,33 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 												{
 													Name:  "RENOVATE_X_GITLAB_AUTO_MERGEABLE_CHECK_ATTEMPS",
 													Value: "7",
+												},
+											},
+										},
+										{
+											Name:  "log-sanitizer",
+											Image: leaktkImageURL,
+											Env: []corev1.EnvVar{
+												{
+													Name:  "HOME",
+													Value: "/tmp",
+												},
+											},
+											Script: logSanitizerScript,
+											SecurityContext: &corev1.SecurityContext{
+												Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+												RunAsNonRoot:             ptr.To(true),
+												RunAsUser:                &normalUser,
+												AllowPrivilegeEscalation: ptr.To(false),
+											},
+											ComputeResources: corev1.ResourceRequirements{
+												Requests: corev1.ResourceList{
+													"cpu":    resource.MustParse("100m"),
+													"memory": resource.MustParse("256Mi"),
+												},
+												Limits: corev1.ResourceList{
+													"cpu":    resource.MustParse("100m"),
+													"memory": resource.MustParse("256Mi"),
 												},
 											},
 										},
--- a/internal/tekton/pipeline_run_builder_test.go
+++ b/internal/tekton/pipeline_run_builder_test.go
@@ -19,11 +19,13 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-multierror"
+	. "github.com/konflux-ci/mintmaker/internal/constant"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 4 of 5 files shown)


internal/tekton/pipeline_run_builder_test.go
```diff
@@ -16,14 +16,17 @@ package tekton
...
 	"time"
 
 	"github.com/hashicorp/go-multierror"
+	. "github.com/konflux-ci/mintmaker/internal/constant"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
```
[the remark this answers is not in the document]
FernandesMF: The imports are divided in two groups: the first has the imports from the standard library, and the second has everything else; the two are separated by a blank line. The imports in each group are sorted correctly.

No task in the CI failed, and `gofmt` does not show any change that would be performed:
```
> gofmt -d internal/tekton/pipeline_run_builder_test.go 
                                                                                                                                            
> 
```

internal/tekton/pipeline_run_builder.go
```diff
@@ -239,6 +243,76 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+										{
+											Name:  "log-sanitizer",
+											Image: leaktkImageURL,
+											Env: []corev1.EnvVar{
+												{
+													Name:  "HOME",
+													Value: "/tmp",
+												},
+											},
+											Script: "#!/bin/sh\n" +
```
staticf0x: Isn't there a better way to embed a script than joining this many lines with escaped characters? It is very difficult to read.

internal/tekton/pipeline_run_builder.go
```diff
@@ -239,6 +243,76 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 												},
 											},
 										},
+										{
+											Name:  "log-sanitizer",
```
staticf0x, line 251: What does this offer on top of Renovate's built-in credential sanitisation?

staticf0x (changes requested):

internal/tekton/pipeline_run_builder.go
FernandesMF: Having this extra step of log sanitization offers two things:
- a backup in case Renovate's sanitization suffers a regression/behaves unexpectedly
- increased coverage

Also, the Renovate docs do [recommend](https://docs.renovatebot.com/security-and-permissions/#log-review-and-redaction-processes) taking scanning and redaction measures in the logs when self-hosting.

#### Backup
Regarding the first point, Renovate's sanitization becoming buggy or suffering a regression is a low-probability but high-impact event.

On one hand, it is unlikely (or at least it should be a rare event). However, the blast radius of that is big, because currently all authenticated users can read all logs.

Because of the big blast radius, it is important to have a mitigation for this risk.

#### Increased coverage
From what I could find in the [docs](https://docs.renovatebot.com/environment-variable-handling/#using-environment-variables-for-secrets), Renovate's sanitization
> relies on knowing whether the value is a secret for instance those configured through secrets or in hostRules.

Thus, if a credential is introduced but it is not configured as one of Renovate's secret, it would not be caught by Renovate's sanitization. 

By using leaktk (and gitleaks underneath), we could catch those credentials that are not configured as secrets in Renovate's env, but still get logged somehow.

staticf0x: Thanks, that's a good reason to have this, I agree

internal/tekton/pipeline_run_builder.go
FernandesMF: This was a great suggestion, thanks!

The script is now extracted to its own file, and it is inserted back using Go embeding (which was a mechanism I didn't know beforehand).

Also, I had unit tests written for the script, which was previously not possible.

internal/constant/constants.go
```diff
@@ -31,4 +31,7 @@ const (
 
 	RenovateImageEnvName    = "RENOVATE_IMAGE"
 	DefaultRenovateImageURL = "quay.io/konflux-ci/mintmaker-renovate-image:latest"
+
+	LeakTKImageEnvName    = "LEAKTK_IMAGE"
+	DefaultLeakTKImageURL = "quay.io/leaktk/leakt:latest"
```
staticf0x:
```suggestion
	DefaultLeakTKImageURL = "quay.io/leaktk/leaktk:latest"
```

internal/tekton/log_sanitizer.sh
```diff
@@ -0,0 +1,49 @@
...
+  echo "$FAIL_MSG"
+  echo "$FAIL_MSG" > "$LOG_FILE"
+  exit 0
+}
+if [ ! -f "$LOG_FILE" ]; then
+  echo 'Log file not found, skipping sanitization'
+  exit 0
+fi
+echo 'Scanning log file for leaked secrets...'
+SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>&1)
```
staticf0x:
```suggestion
SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>/dev/null)
```

With `2>&1`, the output will be:
```
[INFO] queueing scan: id="Ske18OkOmW0" queue_size=1
[INFO] starting scan: id="Ske18OkOmW0"
[INFO] queueing response: id="Ske18OkOmW0" queue_size=1
{"id":"RJY9B0n_COk","kind":"ScanResults","request_id":"Ske18OkOmW0","results":[]}
```

and wouldn't then parse using Python's `json` module, so you need to redirect stderr to `/dev/null` to only get the JSON output

internal/tekton/log_sanitizer.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+if [ -z "$SECRETS" ]; then
+  echo 'No secrets found in log file'
+  exit 0
+fi
+cp "$LOG_FILE" "${LOG_FILE}.tmp"
+echo "$SECRETS" | while IFS= read -r secret; do
+  if [ -n "$secret" ]; then
+    ESCAPED=$(printf '%s\n' "$secret" | sed 's/[[\.*^$()+?{|\\]/\\&/g')
+    sed -i "s|${ESCAPED}|**REDACTED**|g" "${LOG_FILE}.tmp"
```
staticf0x: I've tried to generate a random password including special characters and place it in a real log file and it wasn't replaced in the log file.

The problem is that the previous line `sed 's/[[\.*^$()+?{|\\]/\\&/g'` strips `$` completely, so it won't match for the second sed call.

internal/tekton/log_sanitizer.sh
```diff
@@ -0,0 +1,49 @@
...
+  fail_safe
+fi
+if [ -z "$SECRETS" ]; then
+  echo 'No secrets found in log file'
+  exit 0
+fi
+cp "$LOG_FILE" "${LOG_FILE}.tmp"
+echo "$SECRETS" | while IFS= read -r secret; do
+  if [ -n "$secret" ]; then
+    ESCAPED=$(printf '%s\n' "$secret" | sed 's/[[\.*^$()+?{|\\]/\\&/g')
```
staticf0x: Can you please add a comment detailing what does the sed regex do?

staticf0x (changes requested):

internal/tekton/log_sanitizer.sh
staticf0x: I was actually wrong, I didn't escape it when testing myself.

internal/tekton/pipeline_run_builder.go
```diff
@@ -239,6 +243,76 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+												"echo \"$SECRETS\" | while IFS= read -r secret; do\n" +
+												"  if [ -n \"$secret\" ]; then\n" +
+												"    ESCAPED=$(printf '%s\\n' \"$secret\" | sed 's/[[\\.*^$()+?{|\\\\]/\\\\&/g')\n" +
+												"    sed -i \"s|${ESCAPED}|**REDACTED**|g\" \"${LOG_FILE}.tmp\"\n" +
+												"  fi\n" +
+												"done\n" +
+												"if [ $? -ne 0 ]; then\n" +
+												"  echo 'Failed to redact secrets from log file'\n" +
+												"  fail_safe\n" +
+												"fi\n" +
```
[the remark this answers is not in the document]
FernandesMF: From the three issues pointed:

1. Uses a pipeline into a `while` loop and only checks `$?` after the loop.
2. Uses `sed` with `|` as delimiter but doesn’t escape `|` in the secret.
3. Doesn’t validate the success of key file operations (`cp`, per-replacement `sed`, `mv`)

1 and 3 were indeed bugs and are now addressed.

Issue 2 does not actually happen. There is a new test case to check for that, however.

internal/tekton/pipeline_run_builder.go
```diff
@@ -239,6 +243,76 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+												"if [ ! -f \"$LOG_FILE\" ]; then\n" +
+												"  echo 'Log file not found, skipping sanitization'\n" +
+												"  exit 0\n" +
+												"fi\n" +
+												"echo 'Scanning log file for leaked secrets...'\n" +
+												"SCAN_OUTPUT=$(leaktk scan --kind JSONData \"@${LOG_FILE}\" 2>&1)\n" +
+												"if [ $? -ne 0 ]; then\n" +
+												"  echo \"leaktk scan failed: $SCAN_OUTPUT\"\n" +
+												"  fail_safe\n" +
+												"fi\n" +
```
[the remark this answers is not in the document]
FernandesMF: This issue is now fixed.

internal/tekton/pipeline_run_builder.go
```diff
@@ -239,6 +243,76 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+												"if [ -z \"$SCAN_OUTPUT\" ]; then\n" +
+												"  echo 'No secrets detected or scanner produced no output'\n" +
+												"  exit 0\n" +
+												"fi\n" +
+												"SECRETS=$(echo \"$SCAN_OUTPUT\" | python3 -c \"\nimport sys, json\nfor r in json.load(sys.stdin).get('results',[]):\n    s = r.get('secret','')\n    if s: print(s)\n\")\n" +
+												"if [ $? -ne 0 ]; then\n" +
+												"  echo 'Failed to parse leaktk output'\n" +
+												"  fail_safe\n" +
```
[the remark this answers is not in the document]
FernandesMF: The leaktk image is based off from fedora, which does include python:
https://github.com/leaktk/leaktk/blob/main/Containerfile

I also ran a quick manual check, and that image does have python available:
```
> podman run --rm --entrypoint /bin/sh quay.io/leaktk/leaktk:latest -c "python3 --version" 2>&1
Python 3.14.6
```

internal/tekton/pipeline_run_builder.go
```diff
@@ -239,6 +243,76 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
...
+												"  fi\n" +
+												"done\n" +
+												"if [ $? -ne 0 ]; then\n" +
+												"  echo 'Failed to redact secrets from log file'\n" +
+												"  fail_safe\n" +
+												"fi\n" +
+												"mv \"${LOG_FILE}.tmp\" \"$LOG_FILE\"\n" +
+												"echo 'Log sanitization complete'",
+											SecurityContext: &corev1.SecurityContext{
+												Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
```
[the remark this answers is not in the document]
FernandesMF: This issue was fixed.

internal/constant/constants.go
FernandesMF: Good catch, sorry about that!

internal/tekton/log_sanitizer.sh
FernandesMF: Good catch, thanks a lot! 💡 

I applied the suggestion.

internal/tekton/log_sanitizer.sh
FernandesMF: Comment added. Do you think it clarifies well?

internal/tekton/pipeline_run_builder.go
```diff
@@ -239,6 +247,33 @@ func NewPipelineRunBuilder(name, namespace string) *PipelineRunBuilder {
 												},
 											},
 										},
+										{
+											Name:  "log-sanitizer",
+											Image: leaktkImageURL,
+											Env: []corev1.EnvVar{
+												{
+													Name:  "HOME",
```
staticf0x, line 255: Why do we have to overwrite the `$HOME`?

internal/tekton/log_sanitizer.sh
staticf0x: Yeah it's good, thanks

internal/tekton/pipeline_run_builder.go
FernandesMF: It is to avoid permission restrictions.

The leaktk image would run with a default value of `/` for `HOME`. However, this step will run as an unprivileged user, and writing files there would not be permitted.

The `/tmp` directory, however, is writable by the non-privileged user.

staticf0x (approved):

PR #549, opened 2026-06-24
chore: add golangci-lint and fix reported issues

adelaon:
Adds [golangci-lint](https://golangci-lint.run/) to the repo. Contributors can now run `make lint` locally against a shared config. The linters are now also part of a pre-commit config and a CI workflow.

- **Config** (`.golangci.yaml`): standard linters + `gosec`
- **Makefile**: pin golangci-lint **v2.12.2** (v1.59.1 fails to build on Go 1.26)
- **Docs**: document `make lint` in `AGENTS.md` and document additional linters in CI in `contributing.md`
- **Code**: fix all current lint findings — mostly style/cleanup (dot imports, unused helpers/params in suit test file, embedded metadata fields, defer error handling) plus a few security-minded changes (`gosec` in osv-downloader zip extraction; `nolint` with rationale where behavior is intentional)

No functional behavior changes intended beyond the osv-downloader zip path validation (errors were not read from the function outputs - that is fixed now).

### DIFF
--- a/internal/controller/event_controller.go
+++ b/internal/controller/event_controller.go
@@ -94,7 +94,7 @@ func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 			}
 
 			var plr tektonv1.PipelineRun
-			if err := r.Client.Get(ctx, client.ObjectKey{Namespace: pod.Namespace, Name: plrName}, &plr); err != nil {
+			if err := r.Get(ctx, client.ObjectKey{Namespace: pod.Namespace, Name: plrName}, &plr); err != nil {
 				if apierrors.IsNotFound(err) {
 					// The PipelineRun is gone, we can't update it.
 					return
@@ -118,7 +118,7 @@ func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		}
 	}()
 
-	if err := r.Client.Get(ctx, req.NamespacedName, &evt); err != nil {
+	if err := r.Get(ctx, req.NamespacedName, &evt); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
@@ -144,7 +144,7 @@ func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	podNamespace := evt.InvolvedObject.Namespace
 
 	// Get the actual corresponding Pod object for this event
-	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: podNamespace, Name: podName}, &pod); err != nil {
+	if err := r.Get(ctx, client.ObjectKey{Namespace: podNamespace, Name: podName}, &pod); err != nil {
 		if apierrors.IsNotFound(err) {
 			// Pod has gone, we can't proceed
 			return ctrl.Result{}, nil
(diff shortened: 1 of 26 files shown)


internal/controller/event_controller.go
```diff
@@ -94,7 +94,7 @@ func (r *EventReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 			}
 
 			var plr tektonv1.PipelineRun
-			if err := r.Client.Get(ctx, client.ObjectKey{Namespace: pod.Namespace, Name: plrName}, &plr); err != nil {
+			if err := r.Get(ctx, client.ObjectKey{Namespace: pod.Namespace, Name: plrName}, &plr); err != nil {
```
staticf0x, line 97: Will this respect the cache configured in https://github.com/konflux-ci/mintmaker/blob/main/cmd/manager/main.go#L152 (introduced in https://github.com/konflux-ci/mintmaker/pull/428)?

tools/osv-generator/cve_parser.go
```diff
@@ -31,18 +31,21 @@ func retryGet(url string, maxRetries int, backoff time.Duration) (string, error)
...
-			defer resp.Body.Close()
-			body, err := io.ReadAll(resp.Body)
-			if err != nil {
-				return "", fmt.Errorf("could not read response body: %w", err)
+			body, readErr := io.ReadAll(resp.Body)
+			closeErr := resp.Body.Close()
+			if readErr != nil {
+				return "", fmt.Errorf("could not read response body: %w", readErr)
+			}
+			if closeErr != nil {
```
staticf0x: This changes the logic of this function so that there could potentially be a case, where it successfully read from the response, but couldn't close it properly. Wouldn't we want to continue with the data read in that case?

tools/osv-downloader/download_osv.go
```diff
@@ -67,30 +68,48 @@ func unzipFile(archive_path string, destination string) error {
...
 
+	destination = filepath.Clean(destination)
 	for _, file := range archive.File {
-		filePath := filepath.Join(destination, file.Name)
+		if strings.Contains(file.Name, "..") {
+			return fmt.Errorf("invalid file path in archive: %s", file.Name)
+		}
+		filePath := filepath.Join(destination, file.Name) //nolint:gosec // zip entry paths are validated below
+		cleanPath := filepath.Clean(filePath)
+		if !strings.HasPrefix(cleanPath, destination+string(os.PathSeparator)) && cleanPath != destination {
```
staticf0x, line 80: Can you please add a comment with an example of such invalid path to better illustrate the condition?

staticf0x (changes requested):

internal/controller/event_controller.go
adelaon: yes, those two expressions should be equivalent (`r.Client.Get` and `r.Get`)

iiuc, [the EventReconcile struct directly references the Client](https://github.com/konflux-ci/mintmaker/blob/8bc96710d060d02954e46310d92a56be63bb8c4c/internal/controller/event_controller.go#L39) (I believe that is called [Struct embedding](https://gobyexample.com/struct-embedding)), so the methods and fields can be called directly on the EventReconcile struct

tools/osv-generator/cve_parser.go
adelaon: You are right. The linter had issue with having `defer` within a loop, which is why I added the closing function directly after reading the body, but there is no need to return with a failure when only closing fails - I will fix this.

staticf0x (approved):

PR #557, opened 2026-07-01
chore: add shellcheck, actionlint, and gate fullsend on Go lint

adelaon:
Add hack/shellcheck-embedded-tekton-scripts to extract Tekton Script: strings from pipeline_run_builder.go to be able to run shellcheck on those scripts.

Add a shellcheck to the go-lint CI job (check the shell scripts in PLR builder).

Add actionlint CI check to lint github workflows.

Gate fullsend dispatch on pull_request_target until the Lint Go and PLR shell scripts check passes.

Fix RBAC wildcard grep (grep -Fn) and replace `! -z` with `-n` (these were flagged by the actionlint).

Tests were not implemented for the new `.go` file under hack, because it has no functional effect on the MintMaker controller (it is purely for shellcheck purposes).

Changes are done based on: [CWFHEALTH-5019](https://redhat.atlassian.net/browse/CWFHEALTH-5019)

[CWFHEALTH-5019]: https://redhat.atlassian.net/browse/CWFHEALTH-5019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

edit:
- shellcheck of tekton pipeline `run` scripts was dropped - too high maintanance cost for little benefit
- added renovate custom regex manager to update actionlint version

### DIFF
--- a/.github/workflows/fullsend.yaml
+++ b/.github/workflows/fullsend.yaml
@@ -20,6 +20,7 @@ name: fullsend
 permissions:
   actions: write
   id-token: write
+  checks: read
   contents: read
   pull-requests: read
 
@@ -34,13 +35,36 @@ on:
     types: [submitted]
 
 jobs:
+  wait-for-go-lint:
+    if: >-
+      github.event_name == 'pull_request_target'
+      && contains(fromJSON('["opened","synchronize","ready_for_review"]'), github.event.action)
+    runs-on: ubuntu-latest
+    permissions:
+      checks: read
+      pull-requests: read
+    steps:
+      - name: Wait for Lint Go
+        uses: lewagon/wait-on-check-action@96d9100b431964d10e0136aff8b9ccb92470505e # v1.8.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          check-name: "Lint Go"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          wait-interval: 15
+          checks-discovery-timeout: 600
+
   dispatch:
+    needs: wait-for-go-lint
     concurrency:
       group: fullsend-dispatch-${{ github.event.issue.number || github.event.pull_request.number }}
       cancel-in-progress: false
     if: >-
-      github.event_name != 'issue_comment'
-      || github.event.comment.user.type != 'Bot'
+      always()
+      && (needs.wait-for-go-lint.result == 'success' || needs.wait-for-go-lint.result == 'skipped')
+      && (
+        github.event_name != 'issue_comment'
+        || github.event.comment.user.type != 'Bot'
+      )
     uses: konflux-ci/.fullsend/.github/workflows/dispatch.yml@ec21706cccc58d01588ecd842464a5afcc375ba1 # main
     with:
       event_action: ${{ github.event.action }}
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Check go mod status
         run: |
           go mod tidy
-          if [[ ! -z $(git status -s) ]]
+          if [[ -n $(git status -s) ]]
           then
             echo "Go mod state is not clean:"
             git --no-pager diff
@@ -43,7 +43,7 @@ jobs:
       - name: Check format
         run: |
           make fmt
-          if [[ ! -z $(git status -s) ]]
+          if [[ -n $(git status -s) ]]
           then
             echo "Some files are not properly formatted."
             echo "Please run 'make fmt' and amend your commit."
@@ -53,7 +53,7 @@ jobs:
       - name: Check manifests
         run: |
           make generate manifests
-          if [[ ! -z $(git status -s) ]]
+          if [[ -n $(git status -s) ]]
           then
             echo "Generated sources are not up to date:"
             git --no-pager diff
@@ -61,7 +61,7 @@ jobs:
           fi
       - name: Check RBAC wildcards
         run: |
-          if grep -n "*" config/rbac/*.yaml ; then
+          if grep -Fn '*' config/rbac/*.yaml ; then
             echo "RBAC files contain wildcards (*)"
             exit 1
           fi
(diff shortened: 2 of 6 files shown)


hack/shellcheck-embedded-tekton-scripts/main.go
```diff
@@ -0,0 +1,233 @@
...
+			return true
+		}
+		seen[name] = struct{}{}
+		steps = append(steps, stepScript{name: name, script: script})
+		return true
+	})
+
+	sort.Slice(steps, func(i, j int) bool { return steps[i].name < steps[j].name })
+	return steps, nil
+}
```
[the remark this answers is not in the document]
adelaon: Tests were not implemented for the new .go file under hack, because it has no functional effect on the MintMaker controller (it is purely for shellcheck purposes).

.github/workflows/pr.yml
```diff
@@ -92,6 +99,24 @@ jobs:
...
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Install shellcheck and actionlint
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+          bash <(curl -fsSL https://raw.githubusercontent.com/rhysd/actionlint/v1.7.7/scripts/download-actionlint.bash)
+          sudo mv actionlint /usr/local/bin/
+
```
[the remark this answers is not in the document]
adelaon: Agreed — replaced curl | bash with a pinned release tarball + SHA256 verification from actionlint_1.7.7_checksums.txt; no runtime script fetch anymore.

hack/shellcheck-embedded-tekton-scripts/main.go
```diff
@@ -0,0 +1,233 @@
...
+	var steps []stepScript
+	ast.Inspect(file, func(n ast.Node) bool {
+		comp, ok := n.(*ast.CompositeLit)
+		if !ok {
+			return true
+		}
+		name, script, ok := stepFields(comp)
+		if !ok {
+			return true
+		}
```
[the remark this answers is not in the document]
adelaon: The extractor now fails closed when a composite literal has both Name and Script but either field isn’t a static string (literals / + concat only). It reports file:line and exits non-zero. Unrelated literals (env vars, volume mounts, etc.) are still skipped when they don’t have both fields.

Also added: error on duplicate step names, tests in main_test.go, and a README note on the static-string requirement.

FernandesMF (dismissed):
Looks good and makes sense to me! I think this linting will be very beneficial to the MintMaker project 🚀 

Disclaimer: I am not familiar with AST representation and the Go tooling to parse source code. I did have a look at `hack/shellcheck-embedded-tekton-scripts/main.go` and it seems alright to me, but I probably am not able to catch any bugs or subtleties in the code.

.github/workflows/pr.yml
```diff
@@ -92,6 +99,33 @@ jobs:
...
           config: .kube-linter.yaml
 
+  workflow-lint:
+    name: Lint workflows
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Install shellcheck and actionlint
```
staticf0x, line 102: Why not use the officially recommended way as described in https://github.com/rhysd/actionlint/blob/v1.7.12/docs/usage.md#on-github-actions?

hack/shellcheck-embedded-tekton-scripts/main.go
```diff
@@ -0,0 +1,250 @@
...
+var shellcheckExcludes = []string{
+	"SC2046", // unquoted find in prepare-rpm-cert matches original Tekton script
+}
+
+type stepScript struct {
+	name   string
+	script string
+}
+
+func main() {
```
staticf0x: Is this necessary? Reading AST to extract the few small snippets seems like a lot of work. If it's only for fullsend or something similar, I'd drop this, I worry that maintaining this yields little benefit.

.github/workflows/fullsend.yaml
```diff
@@ -34,13 +35,36 @@ on:
     types: [submitted]
 
 jobs:
+  wait-for-go-lint:
```
staticf0x, line 38: Could this work too?

```yaml
on:
  workflow_run:
    workflows: ["Lint Go and PLR shell scripts"]
    types:
      - completed
```

Docs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_runbranchesbranches-ignore

staticf0x (changes requested):

.github/workflows/pr.yml
adelaon: This way, a pinned release binary from GitHub (v1.7.7) is downloaded and its SHA256 is verified against the official checksum file before it is used (rather then using a directly downloaded script - if that script were compromised, it could execute dangerous code in CI). The recommended way is simpler, but doing it like this is safer. 

I am fine with both approaches, just thought this is more suitable for our use. Do you think using the recommended way from docs would be better here?

staticf0x: That makes sense. How will we update the version and checksum in the future?

hack/shellcheck-embedded-tekton-scripts/main.go
adelaon: This is not only for fullsend (I believe it is in general good to have an extensive CI - for AI agents as well as human developers), but I do agree this script is a lot to maintain. 

There is another option, to extract the scripts into .sh files, add a (way shorter) helper function, that would transform the scripts from files to strings. Then we could use shellcheck on the files directly.

Or, since those scripts are rarely touched, we can just drop this. What do you think is the best approach here?

staticf0x: I think we can drop it altogether. The policy in Konflux mentioned that scanning embedded scripts can be difficult and was considered optional.

.github/workflows/fullsend.yaml
adelaon: There are several points, why I think using this `wait-for-go-lint job` is better:

Firstly, we would need to separate the lint job into it's own workflow (file), since the `workflow_run` works only for workflows and not jobs.

Then to avoid double-dispatch we’d have to drop `opened/synchronize` from `pull_request_target` and dispatch only after CI via `workflow_run`. However, the fullsend dispatch job is using some values from `github.event` and those values would change with changing the trigger: e.g. `github.event.action` becomes completed (not synchronize/opened/…), and the PR object isn’t available the same way (so it would not have the PR context). And for fork PRs, pull_requests can even be empty.

The wait job keeps the real PR event for fullsend every time, and just polls the lint check.

.github/workflows/pr.yml
adelaon: Hmm, good point. The version can be updated by a custom manager in `renovate.json`, but the SHA pin will no longer work.. we can download the checksum file and then verify the actionlint tool download, but I'm not sure there is much benefit - so maybe just keep the pinned version (for reproducibility), updated via renovate, and drop the SHA check altogether?

.github/workflows/fullsend.yaml
staticf0x: Ok yeah sounds good, thank you for the explanation

.github/workflows/pr.yml
staticf0x: Yeah I think so, it'd be the same system we use in https://github.com/konflux-ci/mintmaker-renovate-image/blob/main/Dockerfile, just another dependency. It's not ideal, but we have to choose between manually updating the checksum and having Renovate enabled for this dependency. I'm okay with both, but I slightly tend to automating the updates. We need to make sure we treat the updates as any other PR and review it thoroughly.

staticf0x (approved):

PR #569, opened 2026-07-14
Remove links to migration docs for Konflux tekton tasks

staticf0x:
This has been superseded by having actual release notes. The migrations docs themselves are no longer maintained and they won't work for patch versions (in semver).

This change will use the same table layout like all other PRs, thus unifying the user experience and bringing us closer to the default config.

**Before**
Screenshot From 2026-07-14 13-37-08

**After**
Screenshot From 2026-07-14 13-34-38

cc @chmeliik


KristianTkacik (approved):
LGTM

PR #597, opened 2026-08-07
KONFLUX-9132: renovate: allow update_artifacts_lockfile as postUpgradeTasks command

### LINKED ISSUE #528: Enhancement: Support artifacts.lock.yaml dependencies
For hermetic builds `artifacts.lock.yaml` file is used to fetch external dependencies.
Add support for updating those dependencies.
In case of github releases, the github-{releases,tags} datasource can be used.

Example contents of `artifacts.lock.yaml`:

```
---
metadata: 
  version: "1.0"
artifacts:
  - download_url: "https://github.com/openshift/check-payload/archive/refs/tags/0.3.15.tar.gz"
    checksum: "sha256:83e4f8c203f4cf8cf1056b5e3d0877579f740974eea090c948a04e9548e9a74f"
    filename: "check-payload-0.3.15.tar.gz"
```

https://docs.renovatebot.com/modules/datasource/github-tags/

c4rt0:
Add update_artifacts_lockfile to allowedCommands in config/renovate/self_hosted.json so MintMaker permits repos to call it from postUpgradeTasks.

Without this entry, MintMaker rejects the command even if the script is installed the allowlist is enforced server-side and cannot be overridden per-repo. The regex `^update_artifacts_lockfile .+$` requires a file path argument, preventing the command from being called without a target.

Depends on: https://github.com/konflux-ci/mintmaker-renovate-image/pull/539, the script must be installed in the image before this allowlist entry has any effect.

Closes: https://github.com/konflux-ci/mintmaker/issues/528
Related: https://redhat.atlassian.net/browse/KONFLUX-9132

### DIFF
--- a/config/renovate/renovate.json
+++ b/config/renovate/renovate.json
@@ -642,7 +642,8 @@
   "allowedCommands": [
     "^rpm-lockfile-prototype rpms.in.yaml$",
     "^pipeline-migration-tool migrate -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$",
-    "^refresh-rpm-lockfiles -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$"
+    "^refresh-rpm-lockfiles -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$",
+    "^update_artifacts_lockfile .+$"
   ],
   "updateNotScheduled": false,
   "dependencyDashboard": false,


config/renovate/renovate.json
```diff
@@ -642,7 +642,8 @@
   "allowedCommands": [
     "^rpm-lockfile-prototype rpms.in.yaml$",
     "^pipeline-migration-tool migrate -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$",
-    "^refresh-rpm-lockfiles -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$"
+    "^refresh-rpm-lockfiles -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$",
+    "^update_artifacts_lockfile .+$"
```
rh-rsaar, line 646: This cannot be allowed, this regex captures essential any input, and it'll be easy to use for malicious purposes. The regex must be absolutely strict, no `.+` or `.*` allowed.

rh-rsaar (changes requested):

config/renovate/renovate.json
c4rt0: Fixed in the latest `push --force`

config/renovate/renovate.json
```diff
@@ -238,7 +642,8 @@
   "allowedCommands": [
     "^rpm-lockfile-prototype rpms.in.yaml$",
     "^pipeline-migration-tool migrate -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$",
-    "^refresh-rpm-lockfiles -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$"
+    "^refresh-rpm-lockfiles -f \"\\$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\"$",
+    "^update_artifacts_lockfile ci/hermetic/artifacts\\.lock\\.yaml$"
```
rh-rsaar: This will only work for repos that have the file in this place. Can we make it either:

a) `^update_artifacts_lockfile$` for root, or
b) so that the script also utilises `$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE` and can work on any file that Renovate is updating?

rh-rsaar (changes requested):

config/renovate/renovate.json
c4rt0: Thank's for that suggestion @rh-rsaar.
I think we should go with b). Renovate creates a JSON file `$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE` with metadata about the upgrade, the script reads it to find which artifacts.lock.yaml to process...
I would also need to update our new konflux-ci/update-artifacts-lockfile to accept -f <datafile>. The allowedCommands regex will become generic and any repo with artifacts lockfile can reuse it.

c4rt0: Updated to option b).

rh-rsaar (approved):

rh-rsaar:
Thanks for the work and the patience, merging :)

c4rt0:
I've no permissions, but testing either way:

/ok-to-test

c4rt0:
Ok, fun: I am now blocked from looking at logs at https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/konflux-mintmaker-tenant/pipelinerun/mintmaker-enterprise-contract-6hvbj 
🙅🏻

joelcapitao:
> Ok, fun: I am now blocked from looking at logs at https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/konflux-mintmaker-tenant/pipelinerun/mintmaker-enterprise-contract-6hvbj 🙅🏻

not sure if the EC/Conforma job is mandatory yet

joelcapitao:
> Thanks for the work and the patience, merging :)

@rh-rsaar can we merge it please ? :)

rh-rsaar:
The prefetch-dependencies task was outdated, so I've merged the konflux updates, should hopefully build now

PR #604, opened 2026-08-12
Fix log sanitizer to include error messages in logs

FernandesMF:
Previously, if the leatktk step failed, the specific error message was not included in the logs of that step, which hinders debugging.

This commit changes the logging logic, to make sure the error message is included in the logs, and updates the tests to ensure the correct behavior.

### DIFF
--- a/internal/tekton/log_sanitizer.sh
+++ b/internal/tekton/log_sanitizer.sh
@@ -11,9 +11,11 @@ if [ ! -f "$LOG_FILE" ]; then
   exit 0
 fi
 echo 'Scanning log file for leaked secrets...'
-SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>/dev/null)
+SCAN_STDERR=$(mktemp)
+trap 'rm -f "$SCAN_STDERR"' EXIT
+SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>"$SCAN_STDERR")
 if [ $? -ne 0 ]; then
-  echo 'leaktk scan failed'
+  echo "leaktk scan failed: $(cat "$SCAN_STDERR")"
   fail_safe
 fi
 if [ -z "$SCAN_OUTPUT" ]; then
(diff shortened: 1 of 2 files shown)


internal/tekton/log_sanitizer.sh
```diff
@@ -11,11 +11,14 @@ if [ ! -f "$LOG_FILE" ]; then
   exit 0
 fi
 echo 'Scanning log file for leaked secrets...'
-SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>/dev/null)
+SCAN_STDERR=$(mktemp)
+SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>"$SCAN_STDERR")
 if [ $? -ne 0 ]; then
-  echo 'leaktk scan failed'
+  echo "leaktk scan failed: $(cat "$SCAN_STDERR")"
```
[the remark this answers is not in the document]
FernandesMF: The content of `stderr` from `leaktk` contains operational errors, not scan results. The risk of leaking sensitive information from there is very low.

internal/tekton/log_sanitizer.sh
```diff
@@ -11,11 +11,14 @@ if [ ! -f "$LOG_FILE" ]; then
   exit 0
 fi
 echo 'Scanning log file for leaked secrets...'
-SCAN_OUTPUT=$(leaktk scan --kind JSONData "@${LOG_FILE}" 2>/dev/null)
+SCAN_STDERR=$(mktemp)
```
adelaon, line 14: Instead of writing `rm -f "$SCAN_STDERR"` before every code exit, wouldn't it be better/cleaner to use the `trap` command?
`trap 'rm -f "$SCAN_STDERR"' SIGINT SIGTERM EXIT`

This way we could be certain the temp file is removed on every code exit event (like user interruption, early exit or unexpected issue).

FernandesMF: Great idea!
I switched to using `trap` to remove the temporary stderr file.

I used only the EXIT signal, because it includes all exit modes (SIGINT, SIGTERM, etc).
As for the ERR signal, it is a bash feature, and might not be implemented in all shells, so I didn't include it.

adelaon (approved):

PR #611, opened 2026-08-19
Fix sanitizer handling of input data

FernandesMF:
The sanitization step was failing because it was invoked with the wrong flag, which caused it to fail parsing renovate's NDJSON logs.

This commit changes the invokation flag to "Text", so leaktk treats the input as arbitraty text, and don't expect the input to be a JSON (which wasn't the case, it was actually NDJSON).

There is also a guardrail test to catch regressions of the flag, and a minor adjustment in the test data to reflect renovate's actual data format.

Assisted-by: Claude Opus 4.8


FernandesMF:
I tested the sanitization in stage, and I can confirm it now works as expected:
Screenshot From 2026-08-19 18-12-30
Screenshot From 2026-08-19 18-31-31

adelaon (approved):
LGTM

PR #613, opened 2026-08-20
fix(refresh-rpm-lockfiles): Increase executionTimout for as a temp so…

HozifaWasfy:
…lution

By default renovate has 15 mins timeout for its child processes. Increasing it  to 25 mins as a temporary solution to avoid the SIGTERM signal for refresh-rpm-lockfiles command for repos with >15 dockerfiles. A new solution is being implemented to optimize refresh-rpm-lockfiles. 

Jira:  CWFHEALTH-5092


ElenaKarolinaSemanova (approved):

rh-rsaar (changes requested):
Can you please change the commit message so it doesn't continue in the description?

rh-rsaar (approved):

